PR #56, opened 2024-09-04
feat: add DM-Count parser

jkbmrz:
Adding parser support for [DM-Count](https://github.com/cvlab-stonybrook/DM-Count/tree/master) crowd density estimation model. Moreover, I'm introducing a new message type for 2D numpy arrays of float values. It's used here to represent a density map but could also be re-used elsewhere in the library (e.g. for depth maps but this should go in a separate PR as we might actually also be able to merge the density and depth parsers - this is also why I'm not including any tests in this PR, as I want to have it passed to de-block experiments team and address this in the next PR).

### DIFF
--- a/depthai_nodes/ml/messages/mapping.py
+++ b/depthai_nodes/ml/messages/mapping.py
@@ -1,0 +1,28 @@
+import depthai as dai
+import numpy as np
+
+
+class Map2D(dai.Buffer):
+    def __init__(self):
+        super().__init__()
+        self._map: np.ndarray = None
+        self._width: int = None
+        self._height: int = None
+
+    @property
+    def map(self) -> np.ndarray:
+        return self._map
+
+    @map.setter
+    def map(self, map: np.ndarray):
+        if not isinstance(map, np.ndarray):
+            raise TypeError(
+                f"map must be of type np.ndarray, instead got {type(map)}."
+            )
+        if not len(map.shape) == 2:
+            raise ValueError("array must be a 2D array")
+        if map.dtype != np.float32:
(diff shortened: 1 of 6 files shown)


depthai_nodes/ml/messages/mapping.py
```diff
@@ -0,0 +1,26 @@
+import depthai as dai
+import numpy as np
+
+
+class Map2D(dai.Buffer):
+    def __init__(self):
+        super().__init__()
+        self._map: np.ndarray = None
+        self._width: int = None
+        self._height: int = None
```
kkeroo, line 10: We set _width and _height but we never use them. Should we have getters for those two?

depthai_nodes/ml/parsers/dmcount.py
```diff
@@ -0,0 +1,49 @@
+import depthai as dai
+
+from ..messages.creators import create_density_message
+
+
+class DMCountParser(dai.node.ThreadedHostNode):
```
kkeroo, line 6: This parser looks pretty general. What about renaming it to `DensityParser` or something similar?

kkeroo (approved):
LGTM, just two questions.

depthai_nodes/ml/parsers/dmcount.py
jkbmrz: Yeah that makes sense! However, I'd also merge it together with `MonocularDepthParser` to make it even more general (both parsers output 2d array of float values). So I'd maybe leave it named as it is for now and fix both parsers in the next PR!

klemen1999 (approved):

PR #58, opened 2024-09-09
Adds FastSAM to examples.

kkeroo:
This PR adds FastSAM to the general example. Prompts point and bounding-box are not supported - they require clicking on frame.


jkbmrz (approved):
LGTM

klemen1999 (approved):

PR #59, opened 2024-09-09
Extend SegmentationParser.

kkeroo:
This PR extends the Segmentation parser by also allowing HWC segmentation mask shape - in the case of the DeepLab v3+ model. The parser automatically transposes the tensor.

### DIFF
--- a/depthai_nodes/ml/parsers/segmentation.py
+++ b/depthai_nodes/ml/parsers/segmentation.py
@@ -58,21 +58,23 @@ class SegmentationParser(dai.node.ThreadedHostNode):
             output_layer_names = output.getAllLayerNames()
 
             if len(output_layer_names) != 1:
-                raise ValueError(
-                    f"Expected 1 output layer, got {len(output_layer_names)}."
+                print(
+                    f"Expected 1 output layer, got {len(output_layer_names)}. Will take the first one."
                 )
 
             segmentation_mask = output.getTensor(output_layer_names[0], dequantize=True)
             if len(segmentation_mask.shape) == 4:
                 segmentation_mask = segmentation_mask[0]
-            else:
-                segmentation_mask = segmentation_mask.transpose(2, 0, 1)
 
             if len(segmentation_mask.shape) != 3:
                 raise ValueError(
                     f"Expected 3D output tensor, got {len(segmentation_mask.shape)}D."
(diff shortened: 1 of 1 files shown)


depthai_nodes/ml/parsers/segmentation.py
```diff
@@ -58,21 +58,23 @@ def run(self):
...
 
             if len(segmentation_mask.shape) != 3:
                 raise ValueError(
                     f"Expected 3D output tensor, got {len(segmentation_mask.shape)}D."
                 )
 
+            mask_shape = segmentation_mask.shape
+            min_dim = np.argmin(mask_shape)
+            if min_dim == len(mask_shape) - 1:
+                segmentation_mask = segmentation_mask.transpose(2, 0, 1)
```
jkbmrz, line 77: I'm trying to understand this part:
1. If segmentation_mask is HW/WH, then len(mask_shape) is 2 and min_dim must be 1 in order to make the transpose (which doesn't make sense in this case but I guess this is OK as min_dim will never be equal to 1)
2. If segmentation_mask is HWC/CHW, then len(mask_shape) is 3 and min_dim must be 2 in order to make the transpose. In what case this would happen? Which C dimension expected to be of size 2?

kkeroo: 1. segmentation_mask is always 3D when it comes to this point. The lines above:
```
if len(segmentation_mask.shape) != 
    raise ValueError(
    f"Expected 3D output tensor, got {len(segmentation_mask.shape)}D."
)
```

2. If segmentation_mask is HWC then min_dim is 2 and `len(mask_shape) - 1` is also 2 --> the transpose happens. If segmentation_mask is CHW then min_dim is 0 and `len(mask_shape) - 1` is still 2 --> no transpose.

We simply check which dimension is the smallest. If its the last one we need to transpose it. The assumption here is that the smallest dimension represents classes

jkbmrz: > We simply check which dimension is the smallest. If its the last one we need to transpose it.

Ah I see, `np.argmin` returns the index not the minimal value itself.. Makes sense now!

jkbmrz (approved):

klemen1999 (approved):

PR #60, opened 2024-09-09
Paddle text detection parser

aljazkonec1:
Added Paddle Text detection parser. The output of PaddleOCR Text detection model is a mask of probabilites over the entire image that need to be processed to get valid bounding boxes.


depthai_nodes/ml/parsers/ppocr_text_detection.py
```diff
@@ -0,0 +1,103 @@
...
+        self.max_detections = max_detections
+
+    def setMaskThreshold(self, threshold: float = 0.3):
+        """Sets the mask threshold for creating the mask from model output
+        probabilities.
+
+        @param threshold: The threshold for the mask.
+        @type threshold: float
+        """
+        self.conf_threshold = threshold
```
kkeroo, line 60: `conf_threshold` is not defined. Did you mean `mask_threshold`? Does `mask_threshold` here have the same meaning as `conf_threshold`? If yes, please rename it to `conf_threshold` for consistency.

depthai_nodes/ml/parsers/ppocr_text_detection.py
```diff
@@ -0,0 +1,103 @@
...
+        while self.isRunning():
+            try:
+                output: dai.NNData = self.input.get()
+            except dai.MessageQueue.QueueException:
+                break  # Pipeline was stopped
+
+            layer_names = output.getAllLayerNames()
+            predictions = output.getTensor(layer_names[0], dequantize=True).astype(
+                np.float32
+            )
```
kkeroo, line 89: BTW, there is also function called `getFirstTensor` which gives you the first one

kkeroo (changes requested):
Looks good otherwise, just two small comments. Quick thought: Is it possible to also have ppocr recognition in this parser and then based on incoming tensor shape we run detection decoding or recognition decoding?

depthai_nodes/ml/parsers/ppocr_text_detection.py
aljazkonec1: No, for detection decoding it has to be seperate as its a different model with completely different outputs.

It should be mask_threshold, thanks for noticing. conf_threshold was first then I changed it to mask_threshold.

depthai_nodes/ml/parsers/ppocr_text_detection.py
```diff
@@ -0,0 +1,103 @@
...
+class PPTextDetectionParser(dai.node.ThreadedHostNode):
+    """Parser class for parsing the output of the PP-OCR text detection model.
+
+    Attributes
+    ----------
+    input : Node.Input
+        Node's input. It is a linking point to which the Neural Network's output is linked. It accepts the output of the Neural Network node.
+    out : Node.Output
+        Parser sends the processed network results to this output in a form of DepthAI message. It is a linking point from which the processed network results are retrieved.
+    mask_threshold : float = 0.3
```
jkbmrz, line 17: I'd remove the value hints to keep the docstring consistent with other parsers.

depthai_nodes/ml/parsers/ppocr_text_detection.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(
+        self,
+        mask_threshold: float = 0.3,
+        bbox_threshold: float = 0.7,
+        max_detections: int = 1000,
+    ):
+        """Initializes the PPTextDetectionParser node.
+
+        @param mask_threshold: The threshold for the mask.
+        @type mask_threshold: float = 0.3
```
jkbmrz, line 39: Same comment as above, remove value hints.

jkbmrz (approved):
Some minor syntax comments. LGTM otherwise!

depthai_nodes/ml/parsers/utils/ppocr.py
```diff
@@ -0,0 +1,191 @@
+from typing import Tuple
+
+import cv2
+import numpy as np
+
+
+def _get_mini_boxes(contour: np.ndarray) -> Tuple[np.ndarray, float]:
+    """ "Internal function to get the minimum bounding box of a contour.
```
klemen1999, line 8:
```suggestion
    """ Internal function to get the minimum bounding box of a contour.
```

depthai_nodes/ml/parsers/utils/ppocr.py
```diff
@@ -0,0 +1,191 @@
+from typing import Tuple
```
klemen1999, line 1: Would it make sense to rename this file to `ppdet` or something similar since this utils are not for OCR?

klemen1999 (approved):

depthai_nodes/ml/parsers/utils/ppocr.py
```diff
@@ -0,0 +1,191 @@
...
+
+import cv2
+import numpy as np
+
+
+def _get_mini_boxes(contour: np.ndarray) -> Tuple[np.ndarray, float]:
+    """ Internal function to get the minimum bounding box of a contour.
```
klemen1999, line 8:
```suggestion
    """Internal function to get the minimum bounding box of a contour.
```

depthai_nodes/ml/parsers/utils/ppocr.py
klemen1999: Yes, that works as well

kkeroo (approved):

PR #61, opened 2024-09-10
Lane detection parser.

kkeroo:
This PR adds a Lane Detection parser for Super-fast-lane-detection models. There are two versions of the model so parser requires some parameters. The output of the parser are clusters of points. Each cluster represents one lane. So, for the easiest message representation, I suggest a new message type - Clusters.

PR also adds tests for cluster message and support in general example. MonocularDepthParser is removed from general example because of #57.

### DIFF
--- a/depthai_nodes/ml/messages/__init__.py
+++ b/depthai_nodes/ml/messages/__init__.py
@@ -1,4 +1,5 @@
 from .classification import Classifications
+from .clusters import Cluster, Clusters
 from .img_detections import (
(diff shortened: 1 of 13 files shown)


jkbmrz (approved):
LGTM. I forgot to remove the `setup_monocular_depth_parser` function, my bad (I'm planning to add new support for it now that the https://github.com/luxonis/depthai-nodes/pull/57 is merged).

PR #67, opened 2024-09-13
Vehicle attributes parser

aljazkonec1:
This PR adds a `VehicleAttributesParser` that is used for vehicle-attributes-classification:72x72 . The model has two classification outputs: 
1) Vehicle type, is one of 4 possible types
2) Vehicle color, is one of 7 possible colors

The parser returns a message with attributes `vehicle_type` and `vehicle_color` each being a Tuple like `(class, probability)` of the class with the highest probability.

### DIFF
--- a/depthai_nodes/ml/messages/base_message.py
+++ b/depthai_nodes/ml/messages/base_message.py
@@ -1,0 +1,15 @@
+from typing import Dict, List, Union
+
+import depthai as dai
+
+
+class MiscelaneousMessage(dai.Buffer):
+    def __init__(self):
+        super().__init__()
+        self._data: Dict[str, Union[dai.Buffer, float, List]] = {}
+
+    def setData(self, data: Dict[str, Union[dai.Buffer, float, List]]):
+        self._data = data
+
+    def getData(self) -> Dict[str, Union[dai.Buffer, float, List]]:
+        return self._data
--- a/depthai_nodes/ml/messages/composite.py
+++ b/depthai_nodes/ml/messages/composite.py
@@ -3,7 +3,15 @@ from typing import Dict, List, Union
 import depthai as dai
 
 
-class MiscelaneousMessage(dai.Buffer):
+class CompositeMessage(dai.Buffer):
+    """CompositeMessage class for storing composite of (dai.Buffer, float, List) data.
+
+    Attributes
+    ----------
+    _data : Dict[str, Union[dai.Buffer, float, List]]
+        Dictionary of data with keys as string and values as either dai.Buffer, float or List.
+    """
+
     def __init__(self):
         super().__init__()
         self._data: Dict[str, Union[dai.Buffer, float, List]] = {}
--- a/depthai_nodes/ml/messages/creators/misc.py
+++ b/depthai_nodes/ml/messages/creators/misc.py
@@ -1,6 +1,6 @@
 from typing import List
 
-from ...messages import AgeGender, Classifications
+from ...messages import AgeGender, Classifications, VehicleAttributes
 
 
 def create_age_gender_message(age: float, gender_prob: List[float]) -> AgeGender:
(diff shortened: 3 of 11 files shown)


klemen1999 (commented):
LGTM, I would just like us to think a bit if there is a way to use more general message types instead of custom ones. Because for example `create_vehicle_attributes_message` (and also `VehicleAttributes`) are very model specific. These message are nice from the standpoint of user experience (e.g. can get type just by looking at `.vehicle_type`) but in the longrun this means custom message for every model that is a bit different (e.g. one that predicts brand instead of type).
In general this is a multi-classification model. What if we have a message `MultiClassification` and its data would be a `dict[str, Classification]` (mapping between attribute name and it's Classification - scores and class names)? Or perhaps even more composited message type which has `dict[str, dai.Buffer]` which could be used for this case and also for something like AgeGender?

The drawback of this approach is that these composited types wouldn't have attributes accessable straight on but would have them in a dictionary -> no autocomplete, not that clean,... But I fear that we'll have to create very specific message types for some specific models if we continue down this path.

What are your thoughts on this? Do you have any other possible solutions? (also CC: @kkeroo , @jkbmrz for thoughts)

jkbmrz:
I agree with what @klemen1999 wrote and would go with the second option he proposed:

> ... Or perhaps even more composited message type which has dict[str, dai.Buffer] which could be used for this case and also for something like AgeGender?

I don't think these models will be so common anyways so the drawbacks wouldn't be too annoying and its much nicer to have a clean library than a multitude of custom messages/parsers

aljazkonec1:
I had the same concerns. I can refactor the code maybe as MultiOutputMessage with data dict[string, [dai.Buffer, float, str, np.ndarray] and then we can make subclasses MultiClassification, AgeGender, ..., Anything that has more than one output? Because AgeGender isnt multi classification but one classification one regression.

kkeroo:
I dont like `dict` because of this problem with type hints - it can be really annoying - but we can sacrifice this since we would have example code in the model card. I like your idea @aljazkonec1 with this MultiOutputMessage but I would limit what value types it can handle, e.g. Classifications, Map2D, Keypoints,..., so our "strong" messages? Okay, maybe also float for regression. 

I would also generalize your parser, instead of VehicleAttributesParser you can name it MultiClassifcaitonParser and then map the tensor to the names inside the __init__:

```
tensor_mapping = {"vehicle_types": 0, "vehicle_color": 1}
```
and this can then be set via `setTensorMapping` and defined inside NN archive. These keys can then be used for your dictionary multiclass msg.

aljazkonec1:
Okay I will change accordingly to multiclass and also update nn_archive to incorperate multiple heads.

klemen1999:
> but I would limit what value types it can handle, e.g. Classifications, Map2D, Keypoints,..., so our "strong" messages? Okay, maybe also float for regression.
Agree, since we sacrifice in type hints we should then profit in the lesser amount of custom messages needed (try to keep only strong ones). This MultiOutputMessage could be then built in the parser itself: if we take AgeGender for example parser creates a Classification message for gender and float for age, we package this into MultiOutputMessage and return it (similarly for vehicle model we can have 2 Classification messages packaged together). Does this make sense?

aljazkonec1:
Changed to MultiClassification Parser. 

Also made a Miscellaneous message that only saves a dictionary. The dictionary can only have values dai.Buffer, float or np.ndarray.  To get the dictionary you can call .getData() method thats overwritten from the dai.Buffer class. Therefore at least some autofill is present when using the parser/message.

depthai_nodes/ml/messages/base_message.py
```diff
@@ -0,0 +1,15 @@
+from typing import Dict, List, Union
+
+import depthai as dai
+
+
+class MiscelaneousMessage(dai.Buffer):
```
klemen1999, line 6: Nitpick: I'm not sure about the name. Perhaps `CompositeMessage` (or something similar) fits better since message is composited from other messages?

(in any case I would also rename the file to message name to be consistent with others)

depthai_nodes/ml/messages/creators/misc.py
```diff
@@ -47,3 +50,53 @@ def create_age_gender_message(age: float, gender_prob: List[float]) -> AgeGender
     age_gender_message.gender = gender
 
     return age_gender_message
+
+
+def create_multi_classification_message(
+    classification_attributes: List[str],
+    classification_scores: Union[np.ndarray, List[List[float]]],
+    classification_labels: List[List[str]],
+):
```
klemen1999, line 59: Missing typed function output

depthai_nodes/ml/parsers/vehicle_attributes.py
```diff
@@ -0,0 +1,61 @@
+from typing import List
+
+import depthai as dai
+
+from ..messages.creators import create_multi_classification_message
+
+
+class MultiClassificationParser(dai.node.ThreadedHostNode):
```
klemen1999, line 8: Can we remove any mention of vehicle model in this file - file name, docstrings, variable names - as now this parser should be general for all multi-classification models right?

jkbmrz (approved):
`create_age_gender_message` and `AgeGender` will also need to be changed accordingly but we can do that in a separate PR. LGTM otherwise.

depthai_nodes/ml/parsers/vehicle_attributes.py
```diff
@@ -0,0 +1,61 @@
...
+
+            layer_names = output.getAllLayerNames()
+
+            scores = []
+            for layer_name in layer_names:
+                scores.append(
+                    output.getTensor(layer_name, dequantize=True).flatten().tolist()
+                )
+
+            vehicle_attributes_message = create_multi_classification_message(
```
klemen1999, line 56: Nitpick: Change variables names to not mention vehicles (also the one below this)

klemen1999 (approved):
LGTM

depthai_nodes/ml/messages/creators/misc.py
```diff
@@ -47,3 +50,53 @@ def create_age_gender_message(age: float, gender_prob: List[float]) -> AgeGender
     age_gender_message.gender = gender
 
     return age_gender_message
+
+
+def create_multi_classification_message(
```
kkeroo, line 55: Should we move it to `creators/classification.py` since it is still classification-related msg?

depthai_nodes/ml/messages/composite.py
```diff
@@ -0,0 +1,23 @@
...
+    ----------
+    _data : Dict[str, Union[dai.Buffer, float, List]]
+        Dictionary of data with keys as string and values as either dai.Buffer, float or List.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._data: Dict[str, Union[dai.Buffer, float, List]] = {}
+
+    def setData(self, data: Dict[str, Union[dai.Buffer, float, List]]):
```
kkeroo, line 19: Are type hints strict enough? E.g. we assign Classification msgs inside `create_multi_classification_msg`?

aljazkonec1: I made it dai.Buffer so that we can save multiple forms not just classification messages, but also Keypoints, ImgDetections, XFeat, etc etc. Any of the other major messages

depthai_nodes/ml/messages/creators/misc.py
aljazkonec1: Will move it in next PR where I also update AgeGender to CompositMessage as @jkbmrz  mentioned.

PR #68, opened 2024-09-17
Refactor Code

aljazkonec1:
This PR Refactors parts of the code. Some examples include:
-  moving all `classification` messages into a single file,  
- removing all mentions of vehicles from multiclassification, 
- refactoring `create_age_gender_message` to return CompositeMessage instead of AgeGenderMessage.


kkeroo (approved):
LGTM, did you delete AgeGender msg from `messages`? Did you test that all parsers that are affected work and that examples work?

Not sure if we want to have to parsers in the same file or not (Classification and MultiClassification parsers)? CC @klemen1999 @jkbmrz

aljazkonec1:
Thanks for noticing, yes I forgot to remove AgeGender message. 

All message creators are working, I also tested AgeGender, PaddleOCR, VehicleAttributes parsers. In the example none of these parsers are used so there should be no error.

kkeroo:
AgeGender model is used in examples. Can you adjust the code so it works?

klemen1999 (approved):
LGTM

PR #69, opened 2024-09-17
Generalize MPHandlandmarkParser.

kkeroo:
~~This PR adds a warning to both parsers if the output contains more than one tensor. So, previously we returned `ValueError` but now we will print a warning and take the first tensor. Why? Some Deeplab models have two outputs and only the first one is used. Similarly, Objectron has two outputs - keypoints and score if the obj. is present - I Keypoint parser will only take the first one.~~

~~I think that the score in Objectron is not that important that we should create a separate parser for it because object detector (Yolo or others) should take care that the image contains an object. Thoughts @klemen1999 @jkbmrz?
But it is possible to modify the keypoints parser so that the second output is considered as scores that should be of shape [1] - for object presence - or [n_keypoints, 1] - for every keypoint. Keypoint msg already supports scores.~~

The PR is updated, so we generalize MPHandlandmarkParser into `MPKeypointsParser` which is responsible for all mediapipe keypoints models. It accepts supports 2 or 3 output layers and they need to be named: `keypoints`, `score`, and optionally `handedness`. I am thinking if this can be further merged with Keypoints parser which only has one output tensor - `keypoints` - can address it in the new PR.

New message that extends basic Keypoints msg is added: `KeypointsWithObjectness` - it also has atribute objectness. From this new msg then HandsKeypoints msg is created.

I will fix examples and tests when we settle if this is OK.


depthai_nodes/ml/parsers/keypoints.py
```diff
@@ -53,6 +55,7 @@ def __init__(
 
         self.scale_factor = scale_factor
         self.n_keypoints = n_keypoints
+        self._warned = False
```
jkbmrz, line 58: Should we maybe add this as a parameter (but keep the default value to False)? The added value of doing this is that we have to include it in the docstring and explain what it means (the variable name itself isn't really self explanatory).

jkbmrz (approved):
Added a small suggestion but LGTM otherwise.

jkbmrz:
> I think that the score in Objectron is not that important that we should create a separate parser for it because object detector (Yolo or others) should take care that the image contains an object. Thoughts @klemen1999 @jkbmrz?
> But it is possible to modify the keypoints parser so that the second output is considered as scores that should be of shape [1] - for object presence - or [n_keypoints, 1] - for every keypoint. Keypoint msg already supports scores.

I don't have a strong opinion on this but if support is simple then I'd go for it (avoiding the possible situation where one might require it and is not implemented).

jkbmrz (commented):
LGTM!

> I am thinking if this can be further merged with Keypoints parser which only has one output tensor - keypoints - can address it in the new PR.

I agree, let's make a super-parser for keypoints.

Moreover, we currently have 3 keypoints messages - `Keypoints`, `KeypointsWithObjectness`, `HandKeypoints` and the last two only add a float for objectness and float for handedness. As messages are supposed to be as simple as possible (DAI requirement) I'm thinking if we should do something similar as we did with classification and have such cases covered by the `CompositeMessage`?

PR #71, opened 2024-09-17
[HOT-FIX] Update ppdet to return polygons.

aljazkonec1:
This PR updated ppdet Parser to return a polygons instead of ImgDetections. The polygons are saved as `Keypoints` objects. This is done through the new `create_corner_detection_message` which is also available for other general detection models.

This hotfix does not include tests.

### DIFF
--- a/depthai_nodes/ml/parsers/ppdet.py
+++ b/depthai_nodes/ml/parsers/ppdet.py
@@ -91,7 +91,6 @@ class PPTextDetectionParser(dai.node.ThreadedHostNode):
                 self.max_detections,
             )
 
-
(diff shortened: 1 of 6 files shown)


depthai_nodes/ml/parsers/ppdet.py
```diff
@@ -91,9 +91,9 @@ def run(self):
                 self.max_detections,
             )
 
-            bboxes = corners2xyxy(bboxes)
+            # bboxes = corners2xyxy(bboxes)
```
klemen1999, line 94:
```suggestion

```

klemen1999 (approved):
LGTM

PR #72, opened 2024-09-18
Better error in MPPalmDetectionParser.

kkeroo:
This PR implements better error printing when using wrong scale in MPPalmDetection parser.


jkbmrz (approved):
LGTM

PR #73, opened 2024-09-19
Example of new parser definition

aljazkonec1:
This PR is meant to discuss new parser logic and implementation. The main changes include:
- Change parser parameters to accept a list of `model_heads`  (gotten from nn_archive) and an optional `head_name` parameter to speccify which head to use. If the list of model_heads is == 1, then this parameter is not necessary to be included.

In addition to the change in parsers, the nn_archives and usage sections will need to be updated to reflect the addition of `head_name` argument. This can be done by either adding the head_name parameter in metadata, this requires using `head.metadata.extraParams["name"]`  to get the name, if we have it integrated into the head via dai, we could use `head.name`. 

To get a better feel I also uploaded a private example of the Vehicle attributes model.

Any and all comments/discussion is welcome :)

### DIFF
--- a/depthai_nodes/ml/parsers/classification.py
+++ b/depthai_nodes/ml/parsers/classification.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Dict, List, Union
 
 import depthai as dai
 import numpy as np
@@ -7,9 +7,10 @@ from ..messages.creators import (
     create_classification_message,
     create_multi_classification_message,
 )
+from .parser import Parser
 
 
-class ClassificationParser(dai.node.ThreadedHostNode):
+class ClassificationParser(Parser):
     """Postprocessing logic for Classification model.
 
     Attributes
@@ -31,9 +32,7 @@ class ClassificationParser(dai.node.ThreadedHostNode):
          An object with attributes `classes` and `scores`. `classes` is a list of classes, sorted in descending order of scores. `scores` is a list of corresponding scores.
     """
 
-    def __init__(
-        self, archive_heads: list, head_name: str = "", is_softmax: bool = True
-    ):
+    def __init__(self):
         """Initializes the ClassificationParser node.
 
         Attributes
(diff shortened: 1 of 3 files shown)


aljazkonec1:
In addition if we decide to rewrite the parsers in this manner, that we also refactor them into a base Parser for this [task](https://app.clickup.com/t/86c076mmz) because the ValueErrors in __init__ method will be the same

depthai_nodes/ml/parsers/classification.py
```diff
@@ -31,20 +31,61 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
+                raise ValueError(
+                    "Multiple heads with the same name found in archive, please specify a unique head name"
+                )
+            self.head = head_candidates[0]
+
+        if (
+            self.head.parser != "ClassificationParser"
+        ):  # Maybe we change to a more generic name like 'classification'
+            raise ValueError(
+                "Head is not a classification head, please correct the nn_archive"
```
jkbmrz, line 76: I think this check might be a bit of an overkill.

depthai_nodes/ml/parsers/classification.py
```diff
@@ -31,20 +31,61 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
+    ):
         """Initializes the ClassificationParser node.
 
-        @param classes: List of class names to be used for linking with their respective
-            scores.
-        @param is_softmax: If False, the scores are converted to probabilities using
-            softmax function.
+        Attributes
+        ----------
+        archive_metadata : list
```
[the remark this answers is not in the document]
aljazkonec1: Yes docstrings will be updated when we have decided on final names/adjustments

depthai_nodes/ml/parsers/classification.py
```diff
@@ -31,20 +31,61 @@ class ClassificationParser(dai.node.ThreadedHostNode):
          An object with attributes `classes` and `scores`. `classes` is a list of classes, sorted in descending order of scores. `scores` is a list of corresponding scores.
     """
 
-    def __init__(self, classes: List[str] = None, is_softmax: bool = True):
+    def __init__(
+        self, archive_heads: list, head_name: str = "", is_softmax: bool = True
```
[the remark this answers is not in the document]
aljazkonec1: Yes I agree, also all checks regarding the list size and head_names can be moved into the superclass

depthai_nodes/ml/parsers/classification.py
aljazkonec1: possibly yes, I'm not sure how we will handle integration into dai, if we check for valid parsers when creating the nn_arachive then we can remove this

depthai_nodes/ml/parsers/classification.py
```diff
@@ -31,20 +31,61 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
+
+        layers = self.head.outputs
+        if len(layers) != 1:
+            raise ValueError(
+                "Only one output layer supported for classification, please correct nn_archive/ model"
+            )
+        self.output_layer_name = layers[0]
+
+        self.classes = self.head.metadata.classes
+        self.n_classes = self.head.metadata.nClasses
```
jkbmrz, line 87: should we keep it open that parser-specific parameters could be defined manually? Because as we have it set currently, it's necessary to use a NN Archive. I see how this is useful for DAI but I also see use cases when you'd only have a model and would want to define things manually. In this case, `archive_heads` parameter should be optional.

depthai_nodes/ml/parsers/classification.py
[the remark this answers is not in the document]
jkbmrz: I'd maybe also rename this parameter because its not so intuitive to me what `archive_heads`/`archive_metadata` stands for. Maybe naming it simply as `heads` or `head_metadata` would work?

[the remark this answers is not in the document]
jkbmrz: I see. Then I'd maybe rename this parameter because its not so intuitive to me what `archive_heads`/`archive_metadata` stands for. Maybe naming it simply as `heads` or `head_metadata` would work?

depthai_nodes/ml/parsers/classification.py
kkeroo: IMO its good to check if the parser in NN archive corresponds to the actual parser, because otherwise the user could accidentally put the wrong head_name.

depthai_nodes/ml/parsers/classification.py
kkeroo: I agree that we leave set functions so the user can change the parameters. If we make base parser class we can have a method `build` like dai-experiments host host have it. To the `build` we pass only the head-metadata all that method takes care off mapping all metadata parameters to parser's parameters.

kkeroo (commented):
I think now we can benefit with base parser class with moving all archive_heads and head_name mapping to the base parser. Also, adding `build` method to take care of all parameters

depthai_nodes/ml/parsers/classification.py
aljazkonec1: So the actual parser would have say `metadata: Dict` parameter, and a `.build(metadata: Union(dict, list)` that would do two things: If its a dict it would just set up class attributes if a list (we presume its a list of heads from nn archive) and it assignes class attributes as it does now?

depthai_nodes/ml/parsers/classification.py
jkbmrz: > the user could accidentally put the wrong head_name.

Okay, I see the added value in this

PR #74, opened 2024-09-19
Parser v2 example

aljazkonec1:
This PR adds an updated Parser v2 example. New additions are: 

- A base `Parser` class that all other classes will inherit. 
- Updated sytax using `.build()` to be more in line with dai v3 syntax, the .build method accepts a list of dai.nn_archive.v1.Head objects or a Dictionary. If Head objects are parsed, they are converted to a dictionary.

Docstrings will be updated once the functionality is finalized. Variable names also can be adjusted.

### DIFF
--- a/depthai_nodes/ml/parsers/classification.py
+++ b/depthai_nodes/ml/parsers/classification.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Dict, List, Union
 
 import depthai as dai
 import numpy as np
@@ -7,9 +7,10 @@ from ..messages.creators import (
     create_classification_message,
     create_multi_classification_message,
 )
+from .parser import Parser
 
 
-class ClassificationParser(dai.node.ThreadedHostNode):
+class ClassificationParser(Parser):
     """Postprocessing logic for Classification model.
 
     Attributes
@@ -31,9 +32,7 @@ class ClassificationParser(dai.node.ThreadedHostNode):
          An object with attributes `classes` and `scores`. `classes` is a list of classes, sorted in descending order of scores. `scores` is a list of corresponding scores.
     """
 
-    def __init__(
-        self, archive_heads: list, head_name: str = "", is_softmax: bool = True
-    ):
+    def __init__(self):
         """Initializes the ClassificationParser node.
 
         Attributes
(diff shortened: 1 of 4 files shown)


kkeroo:
I am thinking if we can also move parser-specific parameter setting to build method. E.g. ClassificationParser requires `classes` parameter. Instead of reading it in the parser we can do it inside `build` from BaseParser.

```
class BaseParser:
    def build(self, params: dict):
        # Loop through the dictionary and set attributes
        for key, value in params.items():
            if hasattr(self, key):
                setattr(self, key, value)
            else:
                print(f"Warning: {key} is not a valid attribute")

class ClassificationParser(BaseParser):
    def __init__(self):
        self.param1 = None
        self.param2 = None
        # You can define more default parameters here if needed

# Example usage:
params_dict = {'param1': 'value1', 'param2': 'value2'}

parser = ClassificationParser()
parser.build(params_dict)

print(parser.param1)  # Output: value1
print(parser.param2)  # Output: value2
```

This is quick GPT generated code, but I think it is possible. So, we map all parameters from head->metadata to parser's parameters. If there is no parser parameter we print a warning.

aljazkonec1:
I'm not sure this is going to work because in dai it wil get ported to C, which doesnt support setattr()

depthai_nodes/ml/parsers/parser.py
```diff
@@ -0,0 +1,59 @@
...
+        self.head_configs: Dict = {}
+
+    def build(self, head_metadata: Union[List, Dict], name: str = ""):
+        """Initial build method for all parsers."""
+
+        if isinstance(head_metadata, list):
+            if len(head_metadata) == 0:
+                raise ValueError("No heads parsed from archive")
+            if len(head_metadata) > 1 and name == "":
+                raise ValueError("Multiple heads detected, please specify head name")
```
klemen1999, line 23: Can we make this even smarter: 
If we are initializing ClassificationParser and only one head has `ClassificationParser˙ set as `parser` then we can automatically pick that one even if we get a whole list of all heads passed in.
Something like this:
```
current_parser = self.__class__.__name__ # this should return "ClassificationParser"
all_heads_parsers = [h.parser for h in head_metadata]
num_matches = all_heads_parsers.cound(current_parser)
if num_matches == 1:
    # good, take that one
else:
    # need head name
```

klemen1999 (commented):
The structure LGTM, left one suggestion

depthai_nodes/ml/parsers/parser.py
aljazkonec1: Sounds good, I'll add it.

depthai_nodes/ml/parsers/classification.py
```diff
@@ -71,16 +108,7 @@ def run(self):
...
-            if len(output_layer_names) != 1:
-                raise ValueError(
-                    f"Expected 1 output layer, got {len(output_layer_names)}."
-                )
-
-            if self.n_classes == 0:
-                raise ValueError("Classes must be provided for classification.")
-
-            scores = output.getTensor(output_layer_names[0], dequantize=True).astype(
+            scores = output.getTensor(self.output_layer_name, dequantize=True).astype(
```
jkbmrz, line 111: `self.output_layer_name` is only defined by calling the `build()` method. We are missing a way to define it manually.

jkbmrz (commented):
Adding some more suggestions. Moreover, we are missing docstrings / explanation of parameters. It’s good to add this now so that the reviewers are 100% sure what they look at (cc @aljazkonec1). I'm getting a bit lost in the current parameter naming in the `Parser` class and would rename as follows:
- `head_configs`: Is this a “config” of a single head? If yes, shouldn’t it be named `head_config`? Additionally, not really sure if this is the best name for it.
- `head_metadata`: If I understand correctly this is it metadata of one head or multiple heads (if existent). So I’d rather name it as `heads_metadata` or just `heads` (we iterate through it later in the code and call each iteration a head). Also, if `head_metadata` is a list of single dict, how is this different from the `head_configs`?
- `name`: I’d rename it to `head_name` for better readability.

depthai_nodes/ml/parsers/classification.py
```diff
@@ -31,21 +32,57 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
+            print(
+                "No classes provided in nn_archive metadata. Please provide class names in the nn_archive"
+            )
+
+        try:
+            self.n_classes = self.head_configs["n_classes"]
+        except KeyError:
+            print(
+                "No n_classes provided in nn_archive metadata. Please provide number of classes in the nn_archive"
+            )
```
aljazkonec1, line 82: Are these OK KeyErrors? Or should they be adjusted?

depthai_nodes/ml/parsers/parser.py
```diff
@@ -0,0 +1,90 @@
...
+                num_matches = parser_names_in_archive.count(current_parser)
+                if num_matches == 0:
+                    raise ValueError(
+                        f"No heads available for {current_parser} in the nn_archive."
+                    )
+                elif num_matches == 1:
+                    head = [head for head in heads if head.parser == current_parser][0]
+                else:
+                    raise ValueError(
+                        "Multiple heads detected, please specify a head name."
```
klemen1999, line 52:
```suggestion
                        f"Multiple heads with parser={current_parser} detected, please specify a head name."
```

klemen1999 (approved):
LGTM

depthai_nodes/ml/parsers/classification.py
```diff
@@ -20,32 +21,70 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
+        self.output_layer_name: str = ""
+        self.classes: List = None
+        self.n_classes: int = 0
+        self.is_softmax: bool = True
+
+    def build(
+        self,
+        heads: Union[List, Dict],
+        head_name: str = "",
+        is_softmax: bool = True,
```
kkeroo, line 61: This also should be in nn archive like `classes` and `n_classes`

kkeroo (approved):

depthai_nodes/ml/parsers/classification.py
kkeroo: Hm, This does not error out, right? But just prints out the warning. So, `setClasses` method should be used?

depthai_nodes/ml/parsers/parser.py
```diff
@@ -0,0 +1,90 @@
...
+
+            head_dictionary = {}
+            head_dictionary["parser"] = parser_name
+            head_dictionary["outputs"] = outputs
+            if metadata is not None:
+                head_dictionary.update(metadata.extraParams)
+            self.head_config = head_dictionary
+
+        else:
+            self.head_config = heads
```
jkbmrz, line 88: I'd rather write this as `elif isinstance(heads, dict)` as type hints do not enforce type checking at runtime. So it's possible one provides a heads an int and this code won't return an error. Thus, I'd also add an `else` statement that raises a `ValueError` that the provided heads object is neither a list or a dict.

depthai_nodes/ml/parsers/parser.py
```diff
@@ -0,0 +1,90 @@
...
+
+class Parser(dai.node.ThreadedHostNode):
+    """Base class for all parsers."""
+
+    def __init__(self):
+        super().__init__()
+        self.input = self.createInput()
+        self.out = self.createOutput()
+
+        self.head_config: Dict = {}
```
jkbmrz, line 14: We miss a docstring explanation of the `head_config` parameter.

jkbmrz (commented):
Adding 2 comments to fix in the next PR.

depthai_nodes/ml/parsers/classification.py
```diff
@@ -20,32 +21,57 @@ class ClassificationParser(dai.node.ThreadedHostNode):
...
 
-    def __init__(self, classes: List[str] = None, is_softmax: bool = True):
+    def __init__(self):
         """Initializes the ClassificationParser node.
 
-        @param classes: List of class names to be used for linking with their respective
-            scores.
-        @param is_softmax: If False, the scores are converted to probabilities using
-            softmax function.
+        Attributes
```
tersekmatija, line 38: It makes sense to document that those only get populated after `build()` is called.

PR #75, opened 2024-09-23
Fix RVC4 tensor order

HonzaCuhel:
This PR fixes the issues with RVC4 inference of YOLOv8 kpts, seg and FastSAM models.


kkeroo (approved):
LGTM!

jkbmrz:
I think we should re-open this PR, branching out of `dev` (to many conflicts will arise otherwise).

PR #77, opened 2024-09-24
Parsers-V2: Add RegressionParser

jkbmrz:
Adding a new parser for models with regression output (`RegressionParser`) based on the parsers-v2 logic. I'm also adding a `regression_message_creator` and `Prediction` message.

I've also removed the `AgeGenderParser`. Parsing of the Age-Gender model now consists of two steps. First, we parse age prediction with the `RegressionParser` and, second, we parse the gender prediction with the `ClassificationParser`.


depthai_nodes/ml/messages/creators/regression.py
```diff
@@ -0,0 +1,35 @@
+from typing import List
+
+from ...messages import Prediction, Predictions
+
+
+def create_regression_message(predictions: List[float]) -> Predictions:
+    """Create a DepthAI message for prediction models.
+
+    @param prediction: Predicted value(s).
+    @type prediction: List[float]
```
kkeroo, line 10: docstrings have `prediction` and the function parameter is `predictions`.

kkeroo (approved):
What is the point of having `Predictions` msg - list of `Prediction`? Doesnt the Regression parser parse just one value from the tensor - e.g. age, objectness, handedness?

jkbmrz:
> What is the point of having Predictions msg - list of Prediction? Doesnt the Regression parser parse just one value from the tensor - e.g. age, objectness, handedness?

Good question. The AgeGender model is an example of uni-variate regression (single-value prediction). However, we might also encounter multivariate regression (multi-value prediction) sometime in the future. I don't have an example in mind so I'm not sure how meaningful this is to add at this point.

depthai_nodes/ml/messages/creators/regression.py
jkbmrz: Added in [db7505d](https://github.com/luxonis/depthai-nodes/pull/77/commits/db7505d9111b91c5a64e865f35638979ce3e2f43).

klemen1999 (commented):
LGTM.
I think we can leave the future-proofing for multi-regression heads in. Let's just update the parser to be compatible with new `BaseParser` definiton

PR #78, opened 2024-09-24
Parsers-V2: Refactor Base Parser

jkbmrz:
Adding a small refactor PR for the base parser. Expanding docstrings and fixing the type checking (the issue described [here](https://github.com/luxonis/depthai-nodes/pull/74#discussion_r1773055805)).

Moreover, I'm wondering if:

- the base parser should rather store `head` (depthai object) instead of `head_config` (dict). The functionality would stay the same but this would, in my opinion, make things more clear and readable. It would also simplify the code. I'm assuming the `head_config` is introduced to enable manual parser building (so that one could write a custom dict instead of providing a NN Archive). However, I think that the build method should be used only for building from a NN Archive and we use the parser-specific setter methods for manual building;
- based on the same logic as the previous point, could the `heads` parameter in the build() method be changed from type Union[List, Dict] to type List[depthai.nn_archive.v1.Head]? The Dict option is not necessary if we only allow building from a NN Archive;
- the build() method should have a return statement or can it be omitted? It doesn't really need to return anything;
- we should re-name `Parser` to `BaseParser` and make it an abstract class?

### DIFF
--- a/depthai_nodes/ml/parsers/base_parser.py
+++ b/depthai_nodes/ml/parsers/base_parser.py
@@ -1,5 +1,5 @@
 from abc import ABCMeta, abstractmethod
-from typing import Dict, Union
+from typing import Any, Dict
 
 import depthai as dai
 
@@ -9,7 +9,20 @@ class BaseMeta(ABCMeta, type(dai.node.ThreadedHostNode)):
 
 
 class BaseParser(dai.node.ThreadedHostNode, metaclass=BaseMeta):
-    """Base parser class for neural network output parsers."""
+    """Base class for neural network output parsers. This class serves as a foundation
+    for specific parser implementations used to postprocess the outputs of neural
+    network models. Each parser is attached to a model "head" that governs the parsing
+    process as it contains all the necessary information for the parser to function
+    correctly. Subclasses should implement `build` method to correctly set all
+    parameters of the parser and the `run` method to define the parsing logic.
+
+    Attributes
+    ----------
+    input : Node.Input
+        Node's input. It is a linking point to which the Neural Network's output is linked. It accepts the output of the Neural Network node.
+    out : Node.Output
+        Parser sends the processed network results to this output in a form of DepthAI message. It is a linking point from which the processed network results are retrieved.
+    """
 
     def __init__(self):
         super().__init__()
@@ -25,9 +38,25 @@ class BaseParser(dai.node.ThreadedHostNode, metaclass=BaseMeta):
         pass
 
     @abstractmethod
-    def build(self, head: Union[dai.NNArchive, Dict]):
+    def build(self, head_config: Dict[str, Any]):
+        """Sets the head configuration for the specified head.
+
+        Attributes
+        ----------
+        head_config : Dict
+            A dictionary containing configuration details relevant to the parser, including parameters and settings required for output parsing.
+        """
         pass
 
     @abstractmethod
     def run(self):
+        """Parses the output from the neural network head.
+
+        This method should be overridden by subclasses to implement the specific parsing logic.
+        It accepts arbitrary keyword arguments for flexibility.
+        Args:
+            **kwargs: Arbitrary keyword arguments for the parsing process.
+        Returns:
+            The parsed output message, as defined by the logic in the subclass.
+        """
         pass
--- a/depthai_nodes/ml/parsers/base_parser.py
+++ b/depthai_nodes/ml/parsers/base_parser.py
@@ -1,0 +1,33 @@
+from abc import ABCMeta, abstractmethod
+from typing import Dict, Union
+
+import depthai as dai
+
+
+class BaseMeta(ABCMeta, type(dai.node.ThreadedHostNode)):
+    pass
+
+
+class BaseParser(dai.node.ThreadedHostNode, metaclass=BaseMeta):
+    """Base parser class for neural network output parsers."""
+
+    def __init__(self):
+        super().__init__()
+
+    @property
+    @abstractmethod
+    def input(self) -> dai.Node.Input:
+        pass
+
+    @property
+    @abstractmethod
+    def output(self) -> dai.Node.Output:
+        pass
+
+    @abstractmethod
+    def build(self, head: Union[dai.NNArchive, Dict]):
+        pass
+
+    @abstractmethod
+    def run(self):
+        pass
--- a/depthai_nodes/ml/parsers/classification.py
+++ b/depthai_nodes/ml/parsers/classification.py
@@ -43,7 +43,7 @@ class ClassificationParser(BaseParser):
     ):
         super().__init__()
         self.output_layer_name = output_layer_name
-        self.classes = classes if classes is not None else []
+        self.classes = classes or []
         self.n_classes = n_classes
         self.is_softmax = is_softmax
 
--- a/depthai_nodes/ml/parsers/classification.py
+++ b/depthai_nodes/ml/parsers/classification.py
@@ -1,4 +1,4 @@
-from typing import Dict, List, Union
+from typing import Any, Dict, List
 
 import depthai as dai
 import numpy as np
@@ -7,10 +7,10 @@ from ..messages.creators import (
     create_classification_message,
     create_multi_classification_message,
 )
-from .parser import Parser
+from .base_parser import BaseParser
 
 
-class ClassificationParser(Parser):
+class ClassificationParser(BaseParser):
     """Postprocessing logic for Classification model.
 
     Attributes
@@ -19,6 +19,8 @@ class ClassificationParser(Parser):
         Node's input. It is a linking point to which the Neural Network's output is linked. It accepts the output of the Neural Network node.
     out : Node.Output
         Parser sends the processed network results to this output in a form of DepthAI message. It is a linking point from which the processed network results are retrieved.
+    output_layer_name: str
+        Name of the output layer from which the scores are extracted.
     classes : List[str]
         List of class names to be used for linking with their respective scores. Expected to be in the same order as Neural Network's output. If not provided, the message will only return sorted scores.
     n_classes : int = len(classes)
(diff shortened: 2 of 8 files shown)


aljazkonec1:
- I tried to set type List[dai.nn_archive.v1.Head] but it is not present in python as its ported from C++
- build() has to return self for it to be valid in the usage like pipeline.create(ClassificationParser).build(head_metadata)
- We should make Build accept entire nn_archive as this simplifies the usage for the user as they wont have to write:
```
heads = nn_archive.getConfig().getConfigV1().model.heads
```

depthai_nodes/ml/parsers/parser.py
```diff
@@ -13,17 +27,15 @@ def __init__(self):
 
         self.head_config: Dict = {}
 
-    def build(self, heads: Union[List, Dict], head_name: str = ""):
-        """Initial build method for all parsers. The method sets the head configuration
-        for the parser.
+    def build(self, heads: Union[List, Dict], head_name: str = None):
```
aljazkonec1, line 30: After all the discussion I think its best to change this to
```
def build(self, nn_archive: depthai.NNArchive, head_names: List[str] = None):
```
And add logic to get the metadata from nn_archive + instantiate parsers based on parser name and head name from the metadata.

depthai_nodes/ml/parsers/parser.py
```diff
@@ -1,10 +1,24 @@
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Optional
```
aljazkonec1, line 2: Should we move this parser one folder higher so that the import is nice and more logcal?

depthai_nodes/ml/parsers/parser.py
```diff
@@ -13,78 +27,101 @@ def __init__(self):
 
         self.head_config: Dict = {}
 
-    def build(self, heads: Union[List, Dict], head_name: str = ""):
-        """Initial build method for all parsers. The method sets the head configuration
-        for the parser.
+    def build(self, nn_archive: dai.NNArchive, head_name: str = None):
```
klemen1999, line 30: Let's go with `head_index: int = None` for now. This is optional if we can't automatically match it - in that case we use the list index provided. But if it is specified we don't do any automatic matching and we use that directly. 
If `parser` field in the head from the provided index doesn't match the parser we want to create we also error out so the user knows that either index or parser specified aren't correct

depthai_nodes/ml/parsers/parser.py
klemen1999: I think it's fine to leave this with all other parsers. In most cases users shouldn't import the this base parser class anyway. And we can also add `__all__` in the `__init__` to control "niceness" of the imports - can be addressed down the line.

depthai_nodes/ml/parsers/parser.py
```diff
@@ -1,10 +1,24 @@
-from typing import Dict, List, Union
+from typing import Dict, List, Union, Optional
 
 import depthai as dai
 
 
 class Parser(dai.node.ThreadedHostNode):
```
aljazkonec1, line 6: Should we rename?

klemen1999: We can rename to `BaseParser` yeah to be inline with other libraries (luxonis-train, luxonis-ml)

aljazkonec1:
I've updated the code to the newest version and to be in line with recent depthai NeuralNetworkWithParser done by @dominik737 . 

This update includes the following:
- `class Parser` is now a an interface class to be used as parser instantiater from nn_archive. It accepts nn_archive and an optional head_index parameter and returns a list of instantiated parsers. 
- new class `BaseParser` is an abstract class for all parsers to include
- `ClassificationParser` is now compatible with the new Parser interface. The class attributes can be set in 3 ways:
    1)  during initialization
    2) using `set` methods
    3) using `.build` method that accepts a head saved as a dictionary, this is the preffered way and the way `Parser` class instantiates the parser.

depthai_nodes/ml/parsers/parser.py
```diff
@@ -1,90 +1,57 @@
...
-                        f"Multiple heads with name {head_name} found in nn_archive, please specify a unique name."
-                    )
-                head = head_candidates[0]
+        if head__index:
+            heads = [heads[head__index]]
+        parsers = []
+        pipeline = self.getParentPipeline()
 
+        for head in heads:
             parser_name = head.parser
```
aljazkonec1, line 46: I added indexes before so in case the head_index is given, the key is that index insted of it being 0.

depthai_nodes/ml/parsers/base_parser.py
```diff
@@ -0,0 +1,33 @@
...
+    def input(self) -> dai.Node.Input:
+        pass
+
+    @property
+    @abstractmethod
+    def output(self) -> dai.Node.Output:
+        pass
+
+    @abstractmethod
+    def build(self, head: Union[dai.NNArchive, Dict]):
```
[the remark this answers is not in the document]
jkbmrz: Moreover, if I get it correctly, one can now initialize a parser from a `dai.NNArchive` object only through the `Parser` class which decodes heads and constructs parsers from head configs. So is the type hint `dai.NNArchive` even necessary here?

depthai_nodes/ml/parsers/parser.py
```diff
@@ -1,90 +1,61 @@
...
 
+from . import *
+from .base_parser import BaseParser
+from .utils import decode_head
+
 
 class Parser(dai.node.ThreadedHostNode):
-    """Base class for all parsers."""
+    """General interface for instantiating parsers based on the provided model
+    archive."""
```
jkbmrz, line 12: `Parser` class is now used to initialize parser(s) from a NN Archive. I'd rename it to better indicate its function, e.g. to `ParserInitializer` or `NNArchiveParser` etc.

depthai_nodes/ml/parsers/base_parser.py
```diff
@@ -0,0 +1,33 @@
...
+
+import depthai as dai
+
+
+class BaseMeta(ABCMeta, type(dai.node.ThreadedHostNode)):
+    pass
+
+
+class BaseParser(dai.node.ThreadedHostNode, metaclass=BaseMeta):
+    """Base parser class for neural network output parsers."""
```
jkbmrz, line 12: I think its crucial we add thorough explanation of methods and parameters to the `BaseParser` (they were added in previous commits but are now gone).

jkbmrz (commented):
Adding some comments below.

I have some doubts on the new way of handling the NN Archive parsring. Correct me if wrong, but we lost the ability to initialize a single parser from a NN Archive without using the head index (which is the case for most of the models) as initializing from a NN Archive now always returns a dict of parsers, one for each head. And even if only a single head, a 0 index will need to be used. I'm wondering if we are not over-engineering things here.

depthai_nodes/ml/parsers/classification.py
```diff
@@ -32,44 +34,45 @@ class ClassificationParser(Parser):
...
+    def __init__(
+        self,
+        output_layer_name: str = "",
+        classes: List = None,
+        n_classes: int = 0,
+        is_softmax: bool = True,
+    ):
+        super().__init__()
+        self.output_layer_name = output_layer_name
+        self.classes = classes if classes is not None else []
```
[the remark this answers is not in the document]
aljazkonec1: pre-commit error

depthai_nodes/ml/parsers/base_parser.py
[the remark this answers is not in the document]
aljazkonec1: The Parser has to return a dict of values so that dai team can link it up correctly. If dict has only one key-value pair, it will return ouput as expected. 

Yes you can only initialize it through the parser, this was done in an effort to enforce its usage and to make nn_archive the ground truth / center for all of our apps. Yes sorry I missed a type hint, will fix in next commit.

depthai_nodes/ml/parsers/parser.py
aljazkonec1: I would leave it as is beacause i think its way more readable to instantiate it like this (in a case where NeuralNetworkWithParser isnt used). I would say you remember to instantiate a *Parser* before *ParserInitializer*.

```
pipeline.create(Parser).build(nn_archive)
```

aljazkonec1: @klemen1999 @kkeroo for opinion

depthai_nodes/ml/parsers/parser.py
```diff
@@ -1,90 +1,61 @@
...
-        self.input = self.createInput()
-        self.out = self.createOutput()
-
-        self.head_config: Dict = {}
 
-    def build(self, heads: Union[List, Dict], head_name: str = ""):
-        """Initial build method for all parsers. The method sets the head configuration
-        for the parser.
+    def build(
+        self, nn_archive: dai.NNArchive, head__index: int = None
```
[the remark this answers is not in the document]
aljazkonec1: Mistake

depthai_nodes/ml/parsers/classification.py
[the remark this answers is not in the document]
klemen1999: This should also work
```suggestion
        self.classes = classes or []
```

depthai_nodes/ml/parsers/parser.py
jkbmrz: Maybe also `ParsersGenerator`?

kkeroo: Hm, I would also go with different names (some of the Jakob's proposals)

kkeroo (commented):
I think its fine. I would just change the name `Parser` to something Jakob proposed. Otherwise, my opinion is that we should start with something and iteratively improve it. It seems to me that we are just bouncing things around off the original proposal.

PR #79, opened 2024-09-25
Parsers V2: Refactor KeypointParser

kkeroo:
This PR refactors the Keypoint parser and adds `Keypoints` message as discussed. `Keypoints` msg is a list of `Keypoint` and each `Keypoint` has x,y,z,confidence.


depthai_nodes/ml/parsers/keypoints.py
```diff
@@ -70,6 +89,14 @@ def setNumKeypoints(self, n_keypoints):
         """
         self.n_keypoints = n_keypoints
```
aljazkonec1, line 90: setScaleFactor should set scale_factor not n_keypoints, need to add setNumberKeypoints()

aljazkonec1 (commented):
OK, just add the setNumberKeypoints() method.

depthai_nodes/ml/parsers/keypoints.py
kkeroo: It does. When reviewing some lines that are not affected are hidden :)

aljazkonec1: My bad, the hidden lines matched up  perfectly.

jkbmrz (approved):
LGTM!

PR #82, opened 2024-09-27
XFeat parser for Mono and Stereo.

kkeroo:
This PR changes the way XFeat output is parsed. Currently, we match two consecutive frames from the same input source - not really useful. Now, we introduce two parsers for XFeat - Mono and Stereo versions. Mono version works with one input source (e.g. one color camera) and lets you set the reference frame by triggering with `setTrigger()` method and then all future frames are matched to the reference frame. Stereo version works with two input sources (e.g. left and right camera) and always matches the "left" frame with the "right" frame. 

Additionally, postprocessing is improved - it is faster and uses only bilinear interpolation. Original implementation uses bilinear, bicubic and nearest. This does not affect the final results very much.

This PR is also not relevant to the Parsers V2.


jkbmrz (approved):
LGTM

klemen1999 (approved):

PR #83, opened 2024-09-27
Parsers-V2: Refactor YuNet Parser

jkbmrz:
This PR adds a parsers-v2 refactor for `YuNetParser`, and adjusts `ImgDetectionExtended` message to accept the newly added `Keypoints` object and is also up-to-date with the newly planned changes that will get written in DAI. Detection message creator (`create_detection_message`) is also adjusted and now accepts only numpy arrays (this change also affects `MPPalmDetectionParser`, `SCRFDParser`, and `YOLOExtendedParser` which are not fixed in this PR).

NOTE: The PR also includes some pre-commit format fixing for the Regression parser and message.

### DIFF
--- a/depthai_nodes/ml/parsers/__init__.py
+++ b/depthai_nodes/ml/parsers/__init__.py
@@ -12,7 +12,6 @@ from .mediapipe_palm_detection import MPPalmDetectionParser
 from .mlsd import MLSDParser
 from .parser_generator import ParserGenerator
 from .ppdet import PPTextDetectionParser
-from .classification_sequence import ClassificationSequenceParser
 from .regression import RegressionParser
 from .scrfd import SCRFDParser
 from .segmentation import SegmentationParser
--- a/depthai_nodes/ml/messages/creators/detection.py
+++ b/depthai_nodes/ml/messages/creators/detection.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple, Union
+from typing import List, Union
 
 import depthai as dai
 import numpy as np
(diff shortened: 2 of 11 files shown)


jkbmrz:
One note - we plan to expand `ImgDetections` in DAI to include `x_center`, `y_center`, `height`, `width`, `angle`, `confidence`, `keypoints`, and `mask`. This PR doesn't add `x_center`, `y_center`, `height`, `width` parameters because `xmin`, `ymin`, `xmax`, `ymax` are already implemented and would look a bit weird to have both. Will switch to xywh once DAI implementation is done.

depthai_nodes/ml/parsers/__init__.py
```diff
@@ -12,7 +12,6 @@
 from .mlsd import MLSDParser
 from .parser_generator import ParserGenerator
 from .ppdet import PPTextDetectionParser
-from .classification_sequence import ClassificationSequenceParser
```
kkeroo, line 15: Why removing this?

kkeroo (approved):
LGTM, only one comment

depthai_nodes/ml/parsers/__init__.py
jkbmrz: `from .classification_sequence import ClassificationSequenceParser` is present in line 3. I think this is a consequence of rebranching.

kkeroo: Ah okay, I missed that.

PR #84, opened 2024-09-27
Feat/parsing neural network

dominik737:
Implementation of `ParsingNeuralNetwork` node. The node creates NN node and relevant parser nodes for the `NNArchive`. The task of finding relevant parsers is done by the `ParserGenerator`.


jkbmrz (commented):
LGTM

jkbmrz (approved):

klemen1999 (commented):
LGTM. Please add docstrings to this class as we have for other ones.

kkeroo (approved):
LGTM

depthai_nodes/parsing_neural_network.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def _removeOldParserNodes(self) -> None:
+        for parser in self._parsers.values():
+            self._pipeline.remove(parser)
+
+    def _getParserNodes(self, nnArchive: dai.NNArchive) -> dict[int, BaseParser]:
+        parser_generator = self._pipeline.create(ParserGenerator)
+        parsers = parser_generator.build(nnArchive)
+        for parser in parsers.values():
+            self._nn.out.link(parser.input)
```
tersekmatija, line 37: Can we more specifically define this? (`_nn.out["output_name"].link`)?

depthai_nodes/parsing_neural_network.py
```diff
@@ -0,0 +1,130 @@
+from pathlib import Path
+
+import depthai as dai
+
+from .ml.parsers import BaseParser, ParserGenerator
+
+
+class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
```
tersekmatija, line 8: Does this need to be defined as a host node? I assume it's not possible to just extend the NeuralNetwork node, right?

dominik737: Yea, I tried to inherit from NeuralNetwork, but DepthAI doesn't allow that. I wouldn't view this in bad way thought, because a lot of times it's better to favor composition over inheritance.

depthai_nodes/parsing_neural_network.py
dominik737: I am not sure if I understand what you mean. Right now the only binding in DepthAI is this one:
`    def out(self) -> depthai.Node.Output: ...`

tersekmatija: I believe in v2 at least there was plan to support linking specific outputs from NN nodes. So if your NN had 3 outputs ("cls", "seg", "det"), you could link like:
```
nn.out["cls"].link(cls_parser.input)
nn.out["seg"].link(seg_parser.input)
```
This requires less data to be sent between the nodes. We should check with DAI team if this is in place. If not it's OK as is, but might be worth leaving a TODO here?

dominik737: Okay, I've left a TODO.

klemen1999 (approved):

PR #85, opened 2024-09-30
Fix YuNet parser data types

lnotspotl:
Fix issues with types, as described [here](https://luxonis.slack.com/archives/C065PDYKNAK/p1727215341703719).


kkeroo:
@lnotspotl can you change the PR so it will be merged into `dev` instead of `main`?

PR #90, opened 2024-10-01
 Fix YuNet parser data types

lnotspotl:
This PR fixes the issue with the `YuNet` parser not being able to connect its output to the detection parser's input.

Previously, the output could not be connected to the input because of the way `depthai-core` checks data types send around throughout the pipeline.

Additionally, this PR changes the parent type of the `ImgDetectionsExtended`, allowing `depthai-core` to dynamically cast it to `ImgDetections`. With the parent class being just `dai.Buffer`, `depthai-core` could not cast it and as a result would log an error message.

For more context, refer to [this](https://luxonis.slack.com/archives/C065PDYKNAK/p1727215341703719) conversation.

I removed the last PR as it was rather short and did not want to deal with re-basing.

### DIFF
--- a/depthai_nodes/ml/messages/img_detections.py
+++ b/depthai_nodes/ml/messages/img_detections.py
@@ -99,7 +99,7 @@ class ImgDetectionExtended(dai.ImgDetection):
         self._mask = value
 
 
-class ImgDetectionsExtended(dai.Buffer):
+class ImgDetectionsExtended(dai.ImgDetections):
     """ImgDetectionsExtended class for storing image detections with keypoints.
 
     Attributes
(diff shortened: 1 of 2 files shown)


kkeroo (approved):
LGTM, @lnotspotl is this approach needed also in other custom messages or is it just for ImgDetections?

jkbmrz:
@lnotspotl I'm just working on `YuNetParser` and `ImgDetectionsExtended`, adjusting them for parsers-v2 ([HERE](https://github.com/luxonis/depthai-nodes/pull/83)). If this change can wait a bit more, I can add it there in order to prevent this change get's lost during the merge process?

lnotspotl:
Yeah, that's fine. Add these changes to the linked PR instead. @jkbmrz 

Looking at the implementation, I think we should also change the parent class for the `CornerDetections` class, making it inherit from the `dai.ImgDetections` class.

jkbmrz:
@lnotspotl I think we can close this PR, right?

PR #91, opened 2024-10-01
Parsers-V2: Remove Hand landmark parser and Hand message.

kkeroo:
This PR removes `MPHandLandmarkParser` and `HandKeypoints` message. The parser and the msg were used for hand landmark model. Now, the model consists of 3 heads: Keypoints, score, and handedness. Keypoints output is handled by `KeypointsParser` and the latter two are handled by `RegressionParser`. 

The PR also fixes `RegressionParser` `build` method because it didnt return `self` in the end.

PS: precommit went over the `RegressionParser` so some stylistic changes happened.


jkbmrz (approved):
LGTM

PR #93, opened 2024-10-02
Minor release preparation.

kkeroo:
In this PR we add support in examples for unsupported parsers including:
- MultiClassificationParser
- PaddleOCRParser
- PPTextDetectionParser
- XFeat Mono and Stereo parsers

XFeat example is a bit different so it is not implemented in the general example flow but its rather in its own function.

Next, I extended the MultiClassificationParser because it didnt have set methods. I also increased the tolerance in probability scores validation - if the scores sum to 1. The problem was with efficientnet-lite becuase it predicts 1000 classes so sometimes it didnt meet the criteria (maybe rounding errors or something?).

I adjusted Segmentation and Keypoint parsers so the warning (`Expected 1 output layer, got .... Taking the first one`) is not spammed in real time but rather only printed once.

### DIFF
--- a/depthai_nodes/ml/parsers/keypoints.py
+++ b/depthai_nodes/ml/parsers/keypoints.py
@@ -53,6 +53,7 @@ class KeypointParser(dai.node.ThreadedHostNode):
 
         self.scale_factor = scale_factor
         self.n_keypoints = n_keypoints
+        self._warned = False
 
     def setScaleFactor(self, scale_factor):
         """Sets the scale factor to divide the keypoints by.
@@ -82,10 +83,11 @@ class KeypointParser(dai.node.ThreadedHostNode):
 
             output_layer_names = output.getAllLayerNames()
 
-            if len(output_layer_names) != 1:
-                raise ValueError(
-                    f"Expected 1 output layer, got {len(output_layer_names)}."
+            if len(output_layer_names) != 1 and not self._warned:
+                print(
+                    f"Expected 1 output layer, got {len(output_layer_names)}, will take the first one."
                 )
+                self._warned = True
 
             keypoints = output.getTensor(output_layer_names[0], dequantize=True).astype(
                 np.float32
--- a/depthai_nodes/ml/parsers/segmentation.py
+++ b/depthai_nodes/ml/parsers/segmentation.py
@@ -39,6 +39,7 @@ class SegmentationParser(dai.node.ThreadedHostNode):
         self.input = self.createInput()
         self.out = self.createOutput()
         self.background_class = background_class
+        self._warned = False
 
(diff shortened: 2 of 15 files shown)


examples/utils/xfeat.py
```diff
@@ -0,0 +1,143 @@
...
+                input_shape, type=dai.ImgFrame.Type.BGR888p, fps=fps_limit
+            ),
+            nn_archive,
+        )
+
+        # Set up parser
+        parser = XFeatMonoParser()
+        parser.setOriginalSize(input_shape)
+        parser.setInputSize(input_shape)
+        parser.setMaxKeypoints(2048)
```
aljazkonec1, line 32: Could we, just for the example, reduce the max number of keypoints? When running the example, the visual is not very example friendly. I would rather have maybe max 25 of the most confident keypoints to showcase the model. 
image

aljazkonec1 (commented):
Looks good, added small comment for xfeat example

examples/utils/xfeat.py
kkeroo: Yeah good point. The problem with with less keypoints is that the rectangle can not be accurately calculated. Also here I wanted to showcase that XFeat detects and matches many features compared to others (e.g. SIFT) but true visual is not great. Maybe like you said we display less keypoints in Stereo mode and leave out the rectangle caluation? But in the mono mode I would leave it as it is because when you move the camera fewer kpts will be detected anyway.

aljazkonec1: Yes maybe leave out rectangle for stereo as its either super hard to see with all the keypoints and just showcase keypoints. For mono agree that the num keypoints gets reduceds so we can leave it as-is.

aljazkonec1 (approved):

depthai_nodes/ml/parsers/segmentation.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, background_class=False):
         self.input = self.createInput()
         self.out = self.createOutput()
         self.background_class = background_class
+        self._warned = False
```
jkbmrz, line 42: same as above

depthai_nodes/ml/parsers/keypoints.py
```diff
@@ -53,6 +53,7 @@ def __init__(
 
         self.scale_factor = scale_factor
         self.n_keypoints = n_keypoints
+        self._warned = False
```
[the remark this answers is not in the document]
kkeroo: @jkbmrz I left it out intentionally because it starts with `_`, its private, and not intended to be in interaction with user. Agree?

klemen1999 (approved):

PR #96, opened 2024-10-03
Refactor YOLOExtendedParser to support for all YOLO variants

ptoupas:
This PR modifies the `YOLOExtendedParser` to take into account the use of anchors. The motivation behind that was to support the detection head of the `YOLOP` model.

A _detection only_ mode is also incorporated to the Parser through a flag named `bbox_only` so the parser will only output the detections without keypoints or masks.

A `yolo_version` flag was also added to deal with the special case of `YOLOP` model when calling the [make_grid_numpy()](https://github.com/luxonis/depthai-nodes/blob/6f42c4c716ab7f4d6a67ba6b0e197d6b86c6cb3d/depthai_nodes/ml/parsers/utils/yolo.py#L9-L22) in the [parse_yolo_output()](https://github.com/luxonis/depthai-nodes/blob/6f42c4c716ab7f4d6a67ba6b0e197d6b86c6cb3d/depthai_nodes/ml/parsers/utils/yolo.py#L250-L253) where instead of the `na` it uses the value of `1` for the third argument of the function.


depthai_nodes/ml/parsers/yolo.py
```diff
@@ -66,6 +79,9 @@ def __init__(
         self.iou_threshold = iou_threshold
         self.mask_conf = mask_conf
         self.n_keypoints = n_keypoints
+        self.anchors = anchors if anchors is not None else [None, None, None]
```
klemen1999, line 82:
```suggestion
        self.anchors = anchors or [None, None, None]
```
Should be the same AFAIK

.pre-commit-config.yaml
```diff
@@ -1,3 +1,5 @@
+default_language_version:
```
klemen1999, line 1: Why is this needed?

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -218,25 +237,47 @@ def parse_yolo_output(
...
     @return: Parsed output.
     @rtype: np.ndarray
     """
-    na = 1 if anchors is None else len(anchors)  # number of anchors per head
+    na = 1 if anchors is None else len(anchors) // 2  # number of anchors per head
     bs, _, ny, nx = out.shape  # bs - batch size, ny|nx - y and x of grid cells
 
-    grid = make_grid_numpy(ny, nx, na)
+    if yolo_version is not None and yolo_version.lower() == "yolop":
+        grid = make_grid_numpy(ny, nx, 1)
```
klemen1999, line 251: Can't we also here use `na` instead of specifically 1?

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -173,8 +182,10 @@ def non_max_suppression(
 def parse_yolo_outputs(
     outputs: List[np.ndarray],
     strides: List[int],
-    anchors: np.ndarray,
+    anchors: List[Optional[np.ndarray]],
     kpts: Optional[List[np.ndarray]] = None,
+    det_mode: bool = False,
+    yolo_version: Optional[str] = None,
```
klemen1999, line 188: Is YoloP parsing any different than other Yolo version with anchors (e.g. YoloV5, YoloV4, YoloV3)? Do we need to specify `yolo_version` or is the difference only if it's anchor or anchor-less model (from YoloV6 and newer)?

Basically can we only check if anchors are supplied and then if they are use them, otherwise go with the other postprocessing?

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -146,7 +187,9 @@ def run(self):
                 for o in outputs_names
             ]
 
-            if any("kpt_output" in name for name in layer_names):
+            if self.bbox_only:
```
klemen1999, line 190: Can we set self._DET_MODE=True at the end if it is neither kpts or seg mode (`else` at the end). Because it has to be one of those right? And then we can remove self.bbox_only attribute.

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -66,6 +79,9 @@ def __init__(
         self.iou_threshold = iou_threshold
         self.mask_conf = mask_conf
         self.n_keypoints = n_keypoints
+        self.anchors = anchors if anchors is not None else [None, None, None]
+        self.bbox_only = bbox_only
+        self.yolo_version = yolo_version
```
klemen1999, line 84: Same question as above: Do we need yolo version or can this be implied by looking if anchors are supplied or not?

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -221,15 +266,25 @@ def run(self):
...
                 )
             elif mode == self._SEG_MODE:
                 detections_message = create_detection_message(
-                    np.array(bboxes), np.array(scores), labels, masks=additional_output
+                    bboxes=np.array(bboxes),
+                    scores=np.array(scores),
+                    labels=np.array(labels),
+                    masks=np.array(additional_output),
                 )
+            elif mode == self._DET_MODE:
```
klemen1999, line 281: Can this just be `else`?

klemen1999 (commented):
Thanks, left some comments

jkbmrz (commented):
Adding some comments but generally looks good!

Because this is merging to parser-v2, it should follow the newly defined parser conventions -  `YOLOExtendedParser` should inherit from `BaseParser` and implement `build()` method (see other parsers that were updated recently).

Moreover, I'm wondering if we are not squeezing too many functionalities into a single parser? If we try breaking the parser into utils functions, we could write simple separate parsers for detection, keypoints, and segmentation. Thoughts? @klemen1999 @kkeroo @aljazkonec1

.pre-commit-config.yaml
ptoupas: I added that due to an error I was getting uwing the `pre-commit` with conda environment. I followed the suggestion from this post [here](https://stackoverflow.com/questions/72177222/runtimeerror-failed-to-find-interpreter-for-builtin-discover-of-python-spec-py) and it worked in my case.

depthai_nodes/ml/parsers/utils/yolo.py
ptoupas: No, the `YoloP` is not using `na` value here. It is hardcoded to `1`.
More detailed explanation on [comment](https://github.com/luxonis/depthai-nodes/pull/96#discussion_r1787344543) below.

depthai_nodes/ml/parsers/yolo.py
ptoupas: There might be a problem with this approach. For example the `YoloP` model has both bboxes and masks, so in that case if we don't explicitly define the `bbox_only` flag it will catch the `_masks` in the layer names and the `_SEG_MODE` will be activate. Again this flag was needed (IMO) for dealing with the special case of `YoloP`.

depthai_nodes/ml/parsers/yolo.py
ptoupas: Answered [here](https://github.com/luxonis/depthai-nodes/pull/96#discussion_r1787344543).

depthai_nodes/ml/parsers/yolo.py
ptoupas: Yeap, I am changing this one.

depthai_nodes/ml/parsers/yolo.py
ptoupas: Ok. Adding both changes.

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -35,14 +37,18 @@ class YOLOExtendedParser(dai.node.ThreadedHostNode):
```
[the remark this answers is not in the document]
ptoupas: Adding them.

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -173,8 +182,10 @@ def non_max_suppression(
 def parse_yolo_outputs(
     outputs: List[np.ndarray],
     strides: List[int],
-    anchors: np.ndarray,
+    anchors: List[Optional[np.ndarray]],
```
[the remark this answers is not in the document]
ptoupas: I have made some changes on the type hinting across the code. Please let me know if you are fine with the way I changed that.

klemen1999:
Yes agree, we need to refactor the YoloExtendedParser to be more consice and call smaller, reusable util functions. Ideally this parser should be able to parse YOLO detections and optionally also parse instance keypoints or segmentations if present.

About the versions we are currently syncing but we probably don't need them. In general we should be able to have just 2 groups of Yolos - those with anchors and those without. And on top level both of these can be handled by one parser (like YoloDetectionNetwork works for both in DAI). So let's keep the focus of this PR into adding optional "anchors" (which inherently means adding supoprt for all anchor yolo versions) and then we can refactor the parser itself into smaller parts in the next PR.

depthai_nodes/ml/parsers/yolo.py
klemen1999: This is where ParserV2 comes into play because only layers regarding the detection will be passed into this parser where as layers that should use segmentation parser will not be looked at here.

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -107,10 +187,39 @@ def setNumKeypoints(self, n_keypoints):
...
+        @type yolo_version: YOLOVersion
+        """
+        try:
+            self.yolo_version = YOLOVersion(yolo_version.lower())
+        except ValueError as err:
+            raise ValueError(
+                f"Invalid YOLO version {yolo_version}. Supported YOLO versions are {[e.value for e in YOLOVersion][:-1]}."
+            ) from err
+
+    def setOutputLayerNames(self, output_layer_names):
```
aljazkonec1, line 211: add types to function and what it returns: 
`def setOutputLayerNames(self, output_layer_names: List[str]) -> None:`

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -107,10 +187,39 @@ def setNumKeypoints(self, n_keypoints):
...
 
+    def setAnchors(self, anchors):
+        """Sets the anchors for the YOLO model.
+
+        @param anchors: The anchors for the YOLO model.
+        @type anchors: List[np.ndarray]
+        """
+        self.anchors = anchors
+
+    def setYoloVersion(self, yolo_version):
```
aljazkonec1, line 198: Add types assertions and returns like setOutputLayerNames

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -107,10 +187,39 @@ def setNumKeypoints(self, n_keypoints):
         """
         self.n_keypoints = n_keypoints
 
+    def setAnchors(self, anchors):
```
aljazkonec1, line 190: Same as setOutputLayerNames

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -1,11 +1,21 @@
+from typing import Any, Dict, List, Optional
```
aljazkonec1, line 1: all functions need type hints and return types. See setOutputLayerNames()

depthai_nodes/ml/parsers/yunet.py
```diff
@@ -128,6 +127,32 @@ def setOutputLayerNames(
         self.conf_output_layer_name = conf_output_layer_name
         self.iou_output_layer_name = iou_output_layer_name
 
+    def setInputShape(self, width, height):
```
aljazkonec1, line 130: These two functions are dobled. Probably a mistake in merging changes from other commits?

aljazkonec1 (changes requested):
Logic looks good, just need to add type hints to functions and there are some doubled functions.

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -227,7 +227,7 @@ def keypoints(
         self._keypoints = keypoints
 
 
-class ImgDetectionsExtended(dai.Buffer):
+class ImgDetectionsExtended(dai.ImgDetections):
```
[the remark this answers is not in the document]
ptoupas: Change reverted.

depthai_nodes/ml/parsers/yolo.py
ptoupas: OK, updating all the type hints and return types in functions with the next commit.

depthai_nodes/ml/parsers/yunet.py
ptoupas: Yeap, thanks for noticing that. I am deleting the doubled functions.

depthai_nodes/ml/parsers/utils/bbox.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
```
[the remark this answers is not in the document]
ptoupas: updated the relative functions to the ones at `bbox_format_converters.py`

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -38,6 +59,25 @@ def xywh2xyxy(x: np.ndarray) -> np.ndarray:
     return y
 
 
+def xyxy2xywhn(x: np.ndarray, img_shape: Tuple[int, int]) -> np.ndarray:
```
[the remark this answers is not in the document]
ptoupas: Updated respectively.

depthai_nodes/ml/parsers/yolo.py
ptoupas: done

depthai_nodes/ml/parsers/yolo.py
ptoupas: done

depthai_nodes/ml/parsers/yolo.py
```diff
@@ -331,7 +333,7 @@ def run(self):
                     results[i, 6:],
                 )
 
-                bbox = xyxy2xywhn(bbox, input_shape)
+                bbox = xyxy_to_xywh_norm(bbox, input_shape)
```
aljazkonec1, line 336: Rather use `xyxy_to_xywh(bbox)` and then `normalize_bboxes(bboxes, height, width` so there is no need to make new function in bbox_format_converters.

aljazkonec1 (approved):
Just one small comment left on my part. Otherwise good.

depthai_nodes/ml/parsers/yolo.py
ptoupas: Hi @aljazkonec1 thanks for the comment.
This would not work in this case since we pass a single bbox to the `xyxy_to_xywh_norm` and not a batch of bboxes (shape (4,) instead of (N, 4)).
I could either add an extra dim to the `bbox` before passing to the `xyxy_to_xywh(bbox)` and then `normalize_bboxes` or use this new `xyxy_to_xywh_norm` which is for a single bbox.

aljazkonec1: Is it a problem if we normalize all bboxes after the for loop? I see you use the bbox again in line 351 but the  function aexpects int type bbox, which I'm guessing is it the xyxy format.

Can you check if normalization after for loop and parsing un-normalized bbox to `process_single_mask` causes an error? And if it doesn't I would move normalization out of the loop.

ptoupas: I just updated the code and now we are using the `xyxy_to_xywh` and `normalize_bboxes`. Also I removed completely the `xyxy_to_xywh_norm` function.
The normalization needs to be done in the loop since the `process_single_mask` requires the bbox to be normalized to correctly do the cropping of the mask to the respective bbox.
I believe this workaround (as pushed in the latest commit) should be fine. What do you think @aljazkonec1?

aljazkonec1: Yes if `process_single_mask` needs normalized bbox than this is fine fix. Eventually we would like to also replace `process_single_mask` with just `process_masks` but that will come at a later PR. Looks good!

klemen1999 (commented):
Generally LGTM. Do you mind also @kkeroo  and @jkbmrz giving it a look?

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -173,27 +188,46 @@ def non_max_suppression(
...
-        kpt = kpts[i] if kpts is not None else None
-        out = parse_yolo_output(x, s, a, head_id=i, kpts=kpt)
+    for i, (x, s) in enumerate(zip(outputs, strides)):
+        kpt = kpts[i] if kpts else None
+        a = anchors[i] if anchors else None
+        out = parse_yolo_output(
+            x,
+            s,
+            no,
+            a,
```
jkbmrz, line 225: personally, I'd make the variable names more meaningful. One has to invest time into figuring out what `x`, `s`, `no`, `a` means which is not optimal.

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -276,7 +358,36 @@ def sigmoid(x: np.ndarray) -> np.ndarray:
     return 1 / (1 + np.exp(-x))
 
 
-def process_single_mask(protos: np.ndarray, mask_coeff: np.ndarray, mask_conf: float):
+def crop_mask(mask: np.ndarray, bbox: np.ndarray) -> np.ndarray:
```
jkbmrz, line 361: Should this be moved to some other location in the utils? I don't think this is YOLO specific

jkbmrz (commented):
Two small comments but looking good otherwise!

depthai_nodes/ml/parsers/utils/yolo.py
```diff
@@ -173,27 +189,46 @@ def non_max_suppression(
...
 ) -> np.ndarray:
     """Parse all outputs of an YOLO model (all channels).
 
     @param outputs: List of outputs of an YOLO model.
     @type outputs: List[np.ndarray]
     @param strides: List of strides.
     @type strides: List[int]
-    @param anchors: List of anchors.
-    @type anchors: np.ndarray
+    @param no: Number of outputs of the model.
```
kkeroo, line 204: can we have better variable name? `num_outputs` maybe?

kkeroo (approved):
LGTM! Sorry for late review.

kkeroo:
@klemen1999 will this parser replace native Yolo parser in DAI? Or are we supporting all Yolo versions for future-proof?

depthai_nodes/ml/parsers/utils/yolo.py
ptoupas: Thanks for the suggestion, I have made the change accordingly.
Btw I had just kept that variable name as `no` mostly to be compatible with ultralytics way of naming this variable, but I agree it should be more descriptive.

klemen1999:
> @klemen1999 will this parser replace native Yolo parser in DAI? Or are we supporting all Yolo versions for future-proof?

Yes, the idea is that this would be all-in-one parser for the whole Yolo family of models which would be then ported to DAI and replace the current one implemented as part of the DetectionNetwork node

klemen1999 (approved):

jkbmrz (approved):

PR #100, opened 2024-10-04
Parsers-V2: Refactor Palm detection, SuperAnimal, Lane detection and MLSD

kkeroo:
This PR refactors 4 parsers:
- PalmDetection
- SuperAnimal
- LaneDetection
- M-LSD

All parsers are used for specific models and all models have only one head (no need to change nn archive).

### DIFF
--- a/depthai_nodes/ml/parsers/mediapipe_palm_detection.py
+++ b/depthai_nodes/ml/parsers/mediapipe_palm_detection.py
@@ -1,12 +1,15 @@
+from typing import Any, Dict
+
 import cv2
 import depthai as dai
 import numpy as np
 
 from ..messages.creators import create_detection_message
+from .base_parser import BaseParser
 from .utils import generate_anchors_and_decode
 
 
-class MPPalmDetectionParser(dai.node.ThreadedHostNode):
+class MPPalmDetectionParser(BaseParser):
     """Parser class for parsing the output of the Mediapipe Palm detection model. As the
     result, the node sends out the detected hands in the form of a message containing
     bounding boxes, labels, and confidence scores.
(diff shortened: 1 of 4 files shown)


depthai_nodes/ml/parsers/mediapipe_palm_detection.py
```diff
@@ -48,15 +51,42 @@ def __init__(self, conf_threshold=0.5, iou_threshold=0.5, max_det=100, scale=192
...
+        head_config : Dict
+            The head configuration for the parser.
+
+        Returns
+        -------
+        MPPalmDetectionParser
+            Returns the parser object with the head configuration set.
+        """
+
+        output_layers = head_config["outputs"]
```
aljazkonec1, line 78: Should save to `self.output_layers` attribute and add this attribute in the initializer

depthai_nodes/ml/parsers/mlsd.py
```diff
@@ -45,13 +48,40 @@ def __init__(
...
+        head_config : Dict
+            The head configuration for the parser.
+
+        Returns
+        -------
+        MLSDParser
+            Returns the parser object with the head configuration set.
+        """
+
+        output_layers = head_config["outputs"]
```
aljazkonec1, line 74: Same as for palm detection, should save to `self.output_layers` and add it to initializer and add setMethods.

aljazkonec1 (commented):
Two small changes to add otherwise looks good. 

Just for discussion,  two of the changed parsers are both Keypoint detection and have some duplicate attributes, should we make a functioning `KeypointsParser` and make both of these inherit from it?  So that we dont have so much duplicate code and just override `run` and potentially `build` methods?

depthai_nodes/ml/parsers/mlsd.py
kkeroo: Yeah I left it out intetionally. How is setting `self.output_layers` useful? We still need to get them by actual name e.g. `tpMap` and `heat`.

kkeroo:
>Just for discussion, two of the changed parsers are both Keypoint detection and have some duplicate attributes, should we make a functioning KeypointsParser and make both of these inherit from it? So that we dont have so much duplicate code and just override run and potentially build methods?

Hm, probably we can existing `KeypointParser` as the parent class for `SuperAnimal` parser. Which is the second parser you had in mind? Palm detection is detection parser, MLSD is line detection and LaneDetection is not acctually keypoints?

aljazkonec1:
Yes have `KeypointsParser` be the parent of `SuperAnimal`. Regarding the detection parser maybe there is also a way we can have a general detection parser?

kkeroo:
KeypointParser as parent class to SuperAnimal added in [dd27c83](https://github.com/luxonis/depthai-nodes/pull/100/commits/dd27c8318ea5f989b7675f6059226b1ec18d8031).

aljazkonec1:
Looks great, less code duplication.

depthai_nodes/ml/parsers/mlsd.py
jkbmrz: You could do `self.output_layer_tpmap` and `self.output_layer_heat` and save them separately?

jkbmrz (approved):
LGTM

depthai_nodes/ml/parsers/mlsd.py
klemen1999: Yeah IMO what @jkbmrz suggested makes sense. In the build we check if they are present based on the expected name and if they are we assign them automatically (self.output_layer_tpmap = "tpMap",...) else we error out saying they should use .setOutputLayerTpMap() instead. And in error message we tell them what are the expected output layer names that the model should have if they want to use build().

In run() we then just check if these are set and if they are we take them.

PR #102, opened 2024-10-08
[Refactor] ImgDetectionsExtended and Detection parsers

aljazkonec1:
This PR refactors and/or adds the following:
- `ImgDetectionExtended` is added and is meant for storing rotated bounding boxes of detected objects in the `xc` `yc` `w` `h` `angle` format with optional `keypoints` attribute
-  refactor `create_detection_message` to use the new `ImgDetectionsExtended`
- remove `create_corner_detection_message` and `CornerDetections` as they are now deprecated
- added `DetectionParser` as a general object detection parser that sets common object detection parameters like `conf_threshold``iou_threshold` `max_detections`. Other detection parser then inherit from this class.
- refactor `ppddet` , `yunet`, `scfrd` to parsers-v2 syntax and inheritence structure
- updated `ppdet` postprocessing as it was to restrictive and discarded to many text detections
- aggregated and expanded bbox maniplulations and conversions between bbox formats into a single utils file `bbox_format_converters.py` 
- other small updates and optimizations of some utils functions to be faster and/or compliant with parsers-v2 type assertions and formats.


depthai_nodes/ml/messages/img_detections.py
```diff
@@ -3,16 +3,30 @@
...
 import numpy as np
 
-from .keypoints import Keypoints
+from .keypoints import Keypoint
 
 
-class ImgDetectionExtended(dai.ImgDetection):
-    """ImgDetectionExtended class for storing image detection with keypoints and masks.
+class ImgDetectionExtended:
+    """A class for storing image detections in  with keypoints and masks.
```
jkbmrz, line 10: "in" is a typo

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -3,16 +3,30 @@
...
+        The width of the bounding box.
+    height: float
+        The height of the bounding box.
+    angle: float
+        The angle of the bounding box expressed in degrees.
+    confidence: float
+        Confidence of the detection.
+    label: int
+        Label of the detection.
+    keypoints: Union[List[Tuple[float, float]], List[Tuple[float, float, float]]]
```
jkbmrz, line 28: aren't keypoints of type `List[Keypoint]`?

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -40,44 +60,150 @@ def keypoints(
...
+    def angle(self) -> float:
+        """Returns the angle of the bounding box.
+
+        @return: Angle of the bounding box.
+        @rtype: float
+        """
+        return self._angle
+
+    @x_center.setter
+    def x_center(self, value: float):
```
kkeroo, line 118: Can we check if the value is actually `float` (in all setters)?

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -3,16 +3,30 @@
 import depthai as dai
 import numpy as np
 
-from .keypoints import Keypoints
+from .keypoints import Keypoint
 
 
-class ImgDetectionExtended(dai.ImgDetection):
-    """ImgDetectionExtended class for storing image detection with keypoints and masks.
+class ImgDetectionExtended:
```
jkbmrz, line 9: should this class inherit from `dai.Buffer`?

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -40,44 +60,150 @@ def keypoints(
...
+        """Sets the X coordinate of the center of the bounding box.
+
+        @param value: X coordinate of the center of the bounding box.
+        @type value: float
+        @raise TypeError: If the X coordinate is not a float.
+        """
+        self._x_center = value
+
+    @y_center.setter
+    def y_center(self, value: float):
```
kkeroo, line 128: same as above, and for all setters.

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -21,15 +35,21 @@ class ImgDetectionExtended(dai.ImgDetection):
 
     def __init__(self):
         """Initializes the ImgDetectionExtended object."""
-        super().__init__()
-        self._keypoints: Keypoints = []
-        self._mask: np.ndarray = np.array([])
```
jkbmrz, line 26: If `mask` is removed, why is it still mentioned in the docstring?

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -40,44 +60,150 @@ def keypoints(
...
+        """Sets the label of the detection.
 
-class ImgDetectionsExtended(dai.ImgDetections):
+        @param value: Label of the detection.
+        @type value: int
+        """
+        self._label = value
+
+
+class ImgDetectionsExtended(dai.Buffer):
```
kkeroo, line 206: Does removing `dai.ImgDetections` as a parent class break something? IIRC this was added so some features could work.

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -90,6 +216,7 @@ def __init__(self):
         """Initializes the ImgDetectionsExtended object."""
         super().__init__()
         self._detections: List[ImgDetectionExtended] = []
+        self._masks: np.ndarray = np.array([])
```
kkeroo, line 219: Add to docstrings

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -40,44 +60,150 @@ def keypoints(
...
 
         @param value: List of keypoints.
-        @type value: Keypoints
-        @raise TypeError: If the keypoints are not a Keypoints object.
+        @type value: List[Keypoint]
         """
-        if not isinstance(value, Keypoints):
-            raise TypeError("Keypoints must be a Keypoints object")
-        self._keypoints = value
+        self._keypoints = keypoints
```
jkbmrz, line 70: shouldn't we add some type checking if the value provided to `keypoints` is actually a list of `Keypoint` objects? Type checking is missing also in many other setters!

depthai_nodes/ml/messages/img_detections.py
jkbmrz, line 219: `masks` are not added to the docstring!

depthai_nodes/ml/parsers/detection.py
```diff
@@ -0,0 +1,137 @@
...
+from .base_parser import BaseParser
+
+
+class DetectionParser(BaseParser):
+    def __init__(
+        self,
+        output_layer_name: str = "",
+        conf_threshold: float = 0.5,
+        iou_threshold: float = 0.5,
+        max_detections: int = 100,
```
kkeroo, line 14: Other detection parsers have `max_det` instead of `max_detections`. Stick with `max_det` because this name is used in nn archive

depthai_nodes/ml/parsers/detection.py
```diff
@@ -0,0 +1,137 @@
...
+        Returns
+        -------
+        DetectionParser
+            Returns the parser object with the head configuration set.
+        """
+
+        output_layer = head_config["outputs"]
+        if len(output_layer) != 1:
+            raise ValueError(
+                f"Only one output layer supported for Detection, got {output_layer} layers."
```
kkeroo, line 98: This is not true in many cases: Yunet has more layers, SCRFD has more layers, Palm detection has more layers

depthai_nodes/ml/parsers/detection.py
```diff
@@ -0,0 +1,137 @@
...
+            )
+
+        self.output_layer_names = output_layer[0]
+        self.conf_threshold = head_config.get("conf_threshold", 0.5)
```
jkbmrz, line 102: Could also be written as `self.conf_threshold = head_config.get("conf_threshold", self.conf_threshold)` so if we ever change the initialization value we don't have to change it twice.

depthai_nodes/ml/parsers/scrfd.py
```diff
@@ -90,7 +93,7 @@ def build(
         SCRFDParser
             Returns the parser object with the head configuration set.
         """
-
+        super().build(head_config)
```
kkeroo, line 96: Have you test SCRFD parser? When calling this `build` method from `DetectionParser` it should throw error that `Only one output layer supported for Detection, got {output_layer} layers.`, right?

depthai_nodes/ml/parsers/utils/transform_to_keypoints.py
```diff
@@ -0,0 +1,36 @@
...
+        kp.y = keypoint[1]
+
+        if dim == 3:
+            kp.z = keypoint[2]
+        if confidences:
+            kp.confidence = confidences[i]
+
+        keypoints_list.append(kp)
+
+    return keypoints_list
```
jkbmrz, line 36: could this function be moved to 
`depthai_nodes/ml/parsers/utils/keypoints.py` to keep the amount of utils files low?

depthai_nodes/ml/parsers/utils/bbox_format_converters.py
```diff
@@ -0,0 +1,131 @@
+import numpy as np
+
+
+def corners_to_rotated_bbox(corners: np.ndarray) -> np.ndarray:
```
kkeroo, line 4: can we add simple check that the corners order is correct? E.g. check that first corner is actually top-left (smaller x cordinate than top-right etc.)

depthai_nodes/ml/parsers/yunet.py
```diff
@@ -49,28 +53,24 @@ def __init__(
...
+        @type conf_output_layer_name: str
+        @param iou_output_layer_name: Output layer name for the IoU predictions.
+        @type iou_output_layer_name: str
         """
-        super().__init__()
-        self._out = self.createOutput(
-            possibleDatatypes=[
-                dai.Node.DatatypeHierarchy(dai.DatatypeEnum.ImgDetections, True)
-            ]
-        )
```
jkbmrz, line 58: The `self._out = self.createOutput` part was suggested to be added by [lnotspotl](https://github.com/lnotspotl) #90 to fix the connection between parsers. I'd keep it in!

depthai_nodes/ml/parsers/yunet.py
```diff
@@ -83,6 +83,7 @@ def build(
         YuNetParser
             Returns the parser object with the head configuration set.
         """
+        super().build(head_config)
```
kkeroo, line 86: Same question here as for SCRFD

jkbmrz (approved):
Looking good in general but quite a big PR so some minor improvements need to be made before merging!

jkbmrz (changes requested):
I accidentally clicked approve. Changes need to be made (see above).

kkeroo (changes requested):
Left some comments. Please make sure you test all related models and parsers that are affected

depthai_nodes/ml/messages/img_detections.py
aljazkonec1: When discussing whith dai team, they wanted it to not inherit from dai.ImgDetections. Thats why it's a seperate class.

depthai_nodes/ml/parsers/scrfd.py
aljazkonec1: Yes I did test it but I used it directly and not via ParserGenerator (which uses build) as I thought nn_archives werent updated correctly. The actual parser works, now the build method should also work.

aljazkonec1:
docstrings and type assertions were updated. Smaller bugs were resolved and MP palm detection was additionally converted to parsers v2 format.

jkbmrz:
Looking good! I'm just thinking if we should fully implement what was suggested in #90 (at least until DAI doens't give us a native message). So the part that `ImgDetectionExtended`/`ImgDetectionsExtended` should inherit from `dai.ImgDetection`/`dai.ImgDetections`.

aljazkonec1:
I asked, and they explicitly do not want it to inherit from `dai.ImgDetections`

kkeroo (approved):
LGTM

jkbmrz (approved):

PR #104, opened 2024-10-09
Parsers-V2: Refactor XFeat.

kkeroo:
This PR refactors XFeat parsers. We introduce `XFeatBaseParser` that inherits from `BaseParser` and also defines some common functions. Additionally, `XFeatBaseParser` defines one additional input because stereo version needs two inputs and instead of changing our `BaseParser` I rather implemented it here. Mono and stereo parser classes are merged into one file along with `XFeatBaseParser` class.


depthai_nodes/ml/parsers/xfeat.py
```diff
@@ -0,0 +1,527 @@
...
+            output_layer_heatmaps,
+            original_size,
+            input_size,
+            max_keypoints,
+        )
+
+        self.previous_results = None
+        self.trigger = False
+
+    def build(
```
jkbmrz, line 300: is it even necessary to have the `build()` method here if it adds nothing new to the one defined in the parent class?

depthai_nodes/ml/parsers/xfeat.py
```diff
@@ -0,0 +1,527 @@
...
+        self.max_keypoints = head_config.get("max_keypoints", self.max_keypoints)
+
+        return self
+
+    def setOutputLayerFeats(self, output_layer_feats):
+        """Sets the output layer containing features.
+
+        @param output_layer_feats: Name of the output layer containing features.
+        @type output_layer_feats: str
+        """
```
jkbmrz, line 131: should type checking be added to the setters?

depthai_nodes/ml/parsers/xfeat.py
aljazkonec1, line 300: No need to override the inherited build method here as it doesnt change the behavior of build, just calls the super method.

jkbmrz (approved):
LGTM. Just adding two non-functional comments.

depthai_nodes/ml/parsers/xfeat.py
```diff
@@ -0,0 +1,527 @@
...
+        super().__init__(
+            output_layer_feats,
+            output_layer_keypoints,
+            output_layer_heatmaps,
+            original_size,
+            input_size,
+            max_keypoints,
+        )
+
+    def build(
```
aljazkonec1, line 445: same as for mono, no need

aljazkonec1 (approved):

depthai_nodes/ml/parsers/xfeat.py
kkeroo: Added in [55095b3](https://github.com/luxonis/depthai-nodes/pull/104/commits/55095b38d4b1c80694093331dcb4c303ea22a893).

depthai_nodes/ml/parsers/xfeat.py
kkeroo: Removed in [55095b3](https://github.com/luxonis/depthai-nodes/pull/104/commits/55095b38d4b1c80694093331dcb4c303ea22a893).

kkeroo: Removed in [55095b3](https://github.com/luxonis/depthai-nodes/pull/104/commits/55095b38d4b1c80694093331dcb4c303ea22a893).

depthai_nodes/ml/parsers/xfeat.py
kkeroo: Added in [a87701f](https://github.com/luxonis/depthai-nodes/pull/104/commits/a87701f03b1f6b6e83ec4e3c4b12dd124e3cb1cf). Note that we should do this for all parsers.

PR #110, opened 2024-10-16
fix: clip hrnet scores

jkbmrz:
This is a quick fix for the `HRNetParser`. Scores with calues >1 were occurring (max. observed was 1.1) and this was causing errors later in the process.

### DIFF
--- a/.github/labeler.yaml
+++ b/.github/labeler.yaml
@@ -29,5 +29,3 @@ fix:
     - 'issue/*'
     - 'bugfix/*'
     - 'patch/*'
-release:
-  - base-branch: 'main'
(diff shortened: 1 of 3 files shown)


jkbmrz:
Not really sure what's wrong with the pre-commit check. Running `pre-commit run --all-files` locally spots no issues.

klemen1999 (approved):

kkeroo:
PR is merging to `main`. Are we now switching to the concept @klemen1999 proposed where `main` is not the same as pypi package?

klemen1999:
> PR is merging to main. Are we now switching to the concept @klemen1999 proposed where main is not the same as pypi package?

Yes let's merge to main from now. But we need to be even more sure that we don't introduce breaking changes now

kkeroo (approved):
LGTM. I ran the tests locally and they went through.

PR #112, opened 2024-10-17
Handle special case in Segmentation parser.

kkeroo:
This PR modifies Segmentation Parser. We no longer need to set `background_class` because this is automatically detected out of tensor shapes. I found one bug in previous version. We have a multi-class segmentation model in the ZOO that outputs all classes in one tensor [1, H, W] and not in [n_classes, H, W]. Probably some postprocessing was added to the model (it is Deeplab). This is handled by setting a flag `classes_in_one_layer` and defaults to False. I believe this is rare case, so this paramter is optional and not neccesery to be added to the archive schema.


klemen1999 (approved):
LGTM

jkbmrz (approved):
LGTM.

PR #113, opened 2024-10-21
Feat: Review Parsers v2

jkbmrz:
Adding some final touches before merging the `parsers-v2` to `main`.

**Changes:**
- change `.getConfig().getConfigV1()` calls to `.getConfig()` as introduced in latest DepthAI V3 alpha [releases](https://github.com/luxonis/depthai-core/pull/1133);
- add a quick fix for the `HRNetParser` for scores with values >1 (max. observed was 1.1);
- allow keypoints message creation without setting the `confidence_threshold` paramerer (this way all keypoints are returned);
- unify parser docstrings, remove repeating explanation of `input` and `output` attributes (explained in the `BaseParser`);
- unify order of parser methods to follow the same order as in the `BaseParser`;
- unify setter methods and add missing ones;
- remove `run()` method from the `DetectionParser` because it's not used;
- remove `output_layer_name` parameter from the `DetectionParser` because some parsers that inherit from it parse multiple outputs;
- remove default values from the `DetectionParser` (t might result in mistakes if we don't define them in the child parsers);
- move `score_threshold` parameter setter to `KeypointParser` parent class (not to define it separately in each child class);
- unify how to handle undefined output names within parser `run()` methods;
- unify that in all parser `build()` functions we use `.get()` when calling parameters from `head_config` in order not to output errors if some parameter is missing. **Is this desired? We might want the error out if parameter is not present (but maybe we need a different error at different location)**;
- unify messages and update `List[int]` and `List[float]` objects to numpy arrays;
- rename `SegmentationMasks` message to `SegmentationMasksSAM` and add a new `SegmentationMasks` message. Update `ImgDetectionsExtended` message with the newly added message;
- fix all the existing `unittests` (making them work without big refactoring);
- limit coordinates in `Keypoint` and `ImgDetectionsExtended` messages to 0-1

**TODOs** (could be for next version update):
- unify checking if all necessary parser parameters are defined. We could have an additional `.check()` method for each parser that would be called in the beginning of the `run()` method (e.g. for `FastSamParser` it would be good to check if `yolo_outputs` are set and error out of not);
- in some parsers (e.g. `FastSAMParser` and  `MLSDParser`) we inspect the output shape in order to know if we are dealing with NCHW or NHWC output (depends on the RVC platform). DAI now allows forcing the output format by running `.getTensor( …, dai.TensorInfo.StorageOrder.NCHW,  …)`. We should switch to this where necessary. Additionally, we could also add this to all parsers to make double sure outputs are always drawn in NCHW;
- unify how we define parameters in docstrings of parser methods. Currently, the definition used in the `build()` methods is not the same as in other methods;
- remove `SegmentationMasksSAM` message (use `SegmentationMasks` instead);
- update the `create_segmentation_message` creator to use the new `SegmentationMasks` message (make sure to also update in the `SegmentationParser` that uses the creator);
- review `parsers` (e.g. if parts of the code could me moved to / called from `utils`):
    - `MPPalmDetectionParser`:
        - we should make it utilize `NMS` from the `utils` (instead of importing a custom one from `cv2`);
    -  `YOLOExtendedParser`:
        - should inherit from the `DetectionsParser`;
        - adjust output layer names setting (allow setting it within `__init__()` and `build()` methods) and use them in the run() method (if not defined, error out or write it nicer on how to obtain them);
    - `HRNetParser`:
        - normalize keypoints using utils function instead of doing it manually;
    - `KeypointParser` and `SuperAnimalParser`:
        - utilize some utils function when scaling with the scale factor;
- review `helpers`, `helpers/utils`, `parsers/utils` (this is out of scope of this PR);
- `unittests` need a big refactor (make sure everything is tested, make code more organized, _etc._). Moreover, we need to extend testing to `messages` and `parsers`;
- update `examples`;
- update README.md (e.g. add quick explanations of parser utilization options using setters, `build()`, `ParsingNeuralNetwork`)


kkeroo:
In `ImgDetectionsExtended` we have `masks: SegmentationMask`. This means that we can access the actual mask by going `img_detections_message.masks.mask` this is weird. We could just have `mask: np.array` and access it like `img_detections_message.mask`?

klemen1999:
> In `ImgDetectionsExtended` we have `masks: SegmentationMask`. This means that we can access the actual mask by going `img_detections_message.masks.mask` this is weird. We could just have `mask: np.array` and access it like `img_detections_message.mask`?

Currently this is a bit weird a agree. But having mask as a standalone type allows us to maybe add some data do it down the line if it proves useful. For now I would keep it as is.

And on a different note I would revert the [0ca5313](https://github.com/luxonis/depthai-nodes/pull/113/commits/0ca53136ec15f45eb6c96b5e13b1a4802530497b) commit and allow for bounderies to be [0,1] inclusive so we can have a e.g. detection or keypoint object through the whole image. Or is there an edge case where we don't want this?

Otherwise LGTM

kkeroo:
Okay, but it has bad UX.

> And on a different note I would revert the [0ca5313](https://github.com/luxonis/depthai-nodes/pull/113/commits/0ca53136ec15f45eb6c96b5e13b1a4802530497b) commit and allow for bounderies to be [0,1] inclusive so we can have a e.g. detection or keypoint object through the whole image. Or is there an edge case where we don't want this?

Now we allow that coordinate can be 0 or 1 also. We dont want that? SCRFD sometimes gives coordinate in slightly negative if the bbox is in the edge and when cliping the values to be between 0 and 1 the check passes. If we throw error when `x <= 0` we dont allow such cases

klemen1999:
> Now we allow that coordinate can be 0 or 1 also. We dont want that?

Oh yeah sorry, I misread the code. What we have now is what we want yeah :)

jkbmrz:
cc @kkeroo we could also change the [masks getter](https://github.com/luxonis/depthai-nodes/blob/b0efd4854be44195f48e4addf3d96083480f47ec/depthai_nodes/ml/messages/img_detections.py#L301) so that it would return `self._masks.mask` instead of `self._masks`. This way, calling `img_detections_message.masks` would return the mask array instead of `SegmentationMask` object. What do you think?

kkeroo:
Hm good idea. If @klemen1999 approves do you mind adding it?

klemen1999 (approved):
LGTM, very nice PR guys :rocket:

PR #114, opened 2024-10-22
Parser for Embeddings NN models

ptoupas:
Add a simple parser for embeddings models. Its just passes the output of the NN as an dai.NNData output message

### DIFF
--- a/depthai_nodes/ml/parsers/embeddings.py
+++ b/depthai_nodes/ml/parsers/embeddings.py
@@ -1,0 +1,58 @@
+from typing import Any, Dict
+
+import depthai as dai
+
+from .base_parser import BaseParser
+
+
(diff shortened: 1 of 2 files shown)


depthai_nodes/ml/parsers/embeddings.py
```diff
@@ -0,0 +1,58 @@
...
+            Returns the parser object with the head configuration set.
+        """
+
+        self.output_layer_names = head_config["outputs"]
+        assert (
+            len(self.output_layer_names) == 1
+        ), "Embeddings head should have only one output layer"
+
+        return self
+
```
klemen1999, line 45: Missing setter for output_layer_names (see other parsers)

klemen1999 (approved):
LGTM, let's just add the setter

PR #117, opened 2024-10-25
Examples compatible with parsers v2.

kkeroo:
This PR fixes examples so they work with parsers v2. Almost all models from ZOO do work in examples except multi-head models (Hand landmarker, Objectron, Vehicle attributes, Age-Gender).

Additionally, some small bugs were fixed mainly because we defined too strict checks in messages so some edge cases fail:
- Fix score check: if scores are empty array, we can not check the datatype. Fixed in [ced692c](https://github.com/luxonis/depthai-nodes/pull/117/commits/ced692ca4aa3b0356c35d9c9ed5ec21f5ba79c01).
- Correct mask init: We instantiated mask as np.array instead of `SegmentationMask`. Fixed in [9c9a90c](https://github.com/luxonis/depthai-nodes/pull/117/commits/9c9a90cb901c28b368567f50f3f67382c7e6a8bd).
- Sometimes values do not meet criteria (between 0 and 1) so clipping is done in parsers. Maybe we could just clip the values inside the message itself? Fixed in [611f8cc](https://github.com/luxonis/depthai-nodes/pull/117/commits/611f8cc20c5e50a0ea07934efd6290434d73c107).
- Examples are fixed in [63f879c](https://github.com/luxonis/depthai-nodes/pull/117/commits/63f879c7c8e8cfe5bd578adc627e5449528e1c67).

If the model has more than one head, error is raised.


jkbmrz (approved):
LGTM

PR #121, opened 2024-10-29
Fix on YOLOExtendedParser anchors

ptoupas:
Fix the type and format of anchors in YOLOExtendedParser to be compatible with the format needed in the [NN Archive](https://github.com/luxonis/luxonis-ml/blob/41c0b6b61d71fcf5f96ba0d7cf9e04c307bfbd59/luxonis_ml/nn_archive/config_building_blocks/base_models/head_metadata.py#L58-L61).


kkeroo (approved):
LGTM, please make sure you test it well before merging.

jkbmrz (approved):
Not the biggest fan of the `Optional[List[List[List[float]]]]` notation but I guess we want to keep the NN Archive unchanged? cc @klemen1999

PR #122, opened 2024-11-05
End-to-End test support.

kkeroo:
This PR adds end-to-end testing support in pytest. The most important files are `main.py`, `manual.py`, and `test_e2e.py`. In the test we check the whole process: loading the NN archive, and setting up the neural network nodes, camera nodes, and parsers. Each test is ran as a child process with a timeout set to 30sec. If the model is not supported on the platform or the input size/type is not supported we skip the model. We also skip if the slug is not available in the ZOO etc. 

**Features**:
- Test the pipeline on specific model slug(s)
- Test the pipeline on specific local NN archive(s)
- Test the pipeline on all public ZOO models
- Test the pipeline on all public ZOO models with specific parser (test only one parser)

One can run test locally with `main.py` (pytest) or `manual.py`. See the README for more info.

Downloading models first time takes some time (~4GB). Because of this I made a hardcoded list of all subset of available slugs so we can test the first GH run. I also made a parser-slug mapping dictionary so we dont need to download all models and read their parsers from nn archives every time. This must be optimized - I am talking with hubai dev team about it. Best would be if it parser info could be included in API response.

Github workflow si fired manually. There is an additional parameter we can set. Default: `-all` - means all slugs are tested. We can define any available parameter (e.g. `--parser` or `--slug`). This way we can quickly test open PRs changes.

Simple README is also added, where user can read about testing instructions.


klemen1999 (approved):
LGTM thanks. I think this is a great start which we can build on in next PRs

.github/workflows/e2e_tests.yaml
```diff
@@ -0,0 +1,56 @@
...
+        required: true
+        default: '-all'
+
+jobs:
+  HIL-test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        python-version: [3.12]
```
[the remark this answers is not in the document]
kkeroo: 3.12 is used because hil_framework deps are not supported in 3.8. In the GH runner, python is only used to reserve the testbed and ssh to testbed server. The tests are performed on testbed server.

tests/end_to_end/README.md
```diff
@@ -0,0 +1,43 @@
+# End-to-end tests
+
+This directory contains end-to-end tests for DepthAI. These tests are designed to run on a real device and test the complete pipeline with camera, neural network, and parsers.
+The tests check if the device is able to run a model with neural network node and parse the output with parser node and return the message. If the error is raised during the process, the test will fail.
+
+## Running the tests on GitHub Actions
+
+You can manually trigger the Github action to run the HIL tests. The is a workflow parameter `additional-parameter` that can be used to specify the desired test. Default is `-all`. The available options are: `-all`, `-p <parser_name>`, `-s <slug_1> <slug_2> ...`, `-nn <path_to_archive_1> <path_to_archive_2> ...`.
```
jkbmrz, line 8: "The is a workflow parameter" -> "There is a ..."?

tests/end_to_end/utils.py
```diff
@@ -0,0 +1,152 @@
...
+    """Get the input shape of the model from the NN archive."""
+    inputs = get_inputs_from_archive(nn_archive)
+
+    if len(inputs) > 1:
+        raise ValueError(
+            "This model has more than one input. Currently, only models with one input are supported."
+        )
+
+    try:
+        input_shape = inputs[0].shape[2:][::-1]
```
jkbmrz, line 32: This expects the input shape to be NCHW, right? What if it is NHWC?

jkbmrz (approved):
Two small comments but LGTM otherwise!

tests/end_to_end/README.md
kkeroo: Thanks, fixed in [3ecafa6](https://github.com/luxonis/depthai-nodes/pull/122/commits/3ecafa6cfdb2080d86b0dec9541f8d259f42ba27).

tests/end_to_end/utils.py
kkeroo: It is fixed in [124e081](https://github.com/luxonis/depthai-nodes/pull/122/commits/124e081c85bab415f408068567b086312b21bf9e).

PR #123, opened 2024-11-06
feat: add version to pyproject.toml

jkbmrz:
Adding a simple fix to `pyproject.toml` to fix the missing version in `pypi` manifest.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,6 @@
 [project]
 name = "depthai-nodes"
+version = "0.1.0"
 description = "Community on-host nodes for DepthAI."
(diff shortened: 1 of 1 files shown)


kkeroo (approved):

klemen1999:
Let's rather not merge this in until we are sure that it solves the issue because if we have it like that it is annoying when we make a new release. We need to explictily change version in two places: `__init__.py` and in `pyproject.toml` whereas currently we only define it in init and it gets dynamically populated in pyproject.toml

jkbmrz:
We made to make it work without this change!

PR #124, opened 2024-11-06
Message improvements (properties and methods).

kkeroo:
This PR improves 3 messages:

1. `Predictions` msg - we add a property `prediction` for retrieving first prediction value. This is useful because Regression parser many times returns one value and to get the value we must type `message.predictions[0].prediction`. Now, we will just write `message.prediction`.
2. `Classification` msg - same idea as with Predictions msg. to get the most probable class and score we add two properties: `top_class` and `top_score`. 
3. `ImgDetectionExtended` msg - two methods are added for easy conversion to [x1,y1,x2,y2] bbox and for [[x1,y1], [x2,y2],[x3,y3],[x4,y4]] rotated bbox (going from the top-left and clockwise)

### DIFF
--- a/depthai_nodes/ml/messages/classification.py
+++ b/depthai_nodes/ml/messages/classification.py
@@ -72,3 +72,21 @@ class Classifications(dai.Buffer):
         if value.size > 0 and value.dtype != np.float32:
             raise ValueError("Scores must be a np.ndarray of floats.")
         self._scores = value
+
+    @property
+    def top_class(self) -> str:
+        """Returns the most probable class.
+
+        @return: The top class.
+        @rtype: str
+        """
+        return self._classes[0]
+
+    @property
+    def top_score(self) -> float:
+        """Returns the probability of the most probable class.
+
+        @return: The top score.
+        @rtype: float
+        """
+        return self._scores[0]
--- a/depthai_nodes/ml/messages/img_detections.py
+++ b/depthai_nodes/ml/messages/img_detections.py
@@ -241,7 +241,8 @@ class ImgDetectionExtended(dai.Buffer):
         self._keypoints = value
 
     def get_xyxy_bbox_points(self) -> NDArray[np.float32]:
-        """Returns the [x1, y1, x2, y2] bounding box points.
+        """Returns the axis-aligned [x1, y1, x2, y2] bounding box points. It does not
+        take into account the angle of the bounding box.
 
         @return: Bounding box points.
         @rtype: NDArray[np.float32]
--- a/depthai_nodes/ml/messages/img_detections.py
+++ b/depthai_nodes/ml/messages/img_detections.py
@@ -240,6 +240,44 @@ class ImgDetectionExtended(dai.Buffer):
             raise ValueError("Keypoints must be a list of Keypoint objects.")
         self._keypoints = value
 
+    def get_xyxy_bbox_points(self) -> NDArray[np.float32]:
+        """Returns the [x1, y1, x2, y2] bounding box points.
+
+        @return: Bounding box points.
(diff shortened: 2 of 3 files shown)


depthai_nodes/ml/messages/img_detections.py
```diff
@@ -240,6 +240,44 @@ def keypoints(
             raise ValueError("Keypoints must be a list of Keypoint objects.")
         self._keypoints = value
 
+    def get_xyxy_bbox_points(self) -> NDArray[np.float32]:
+        """Returns the [x1, y1, x2, y2] bounding box points.
```
klemen1999, line 244: If bbox is rotated this gives axis-alligned bbox right? If so maybe we want to explicitly mention "axis-alligned bounding box" in the docstring

klemen1999 (approved):
LGTM

depthai_nodes/ml/messages/img_detections.py
kkeroo: Yes true, it doesnt take into account the angle. Updated docstring in [c90782e](https://github.com/luxonis/depthai-nodes/pull/124/commits/c90782e683de48376b09392de1ad62db10e3bab4).

jkbmrz (approved):
LGTM. The added features seems useful! I'm wondering, though, if we are not making the messages too complicated for DAI to implement natively (e.g. for bbox-related methods, wouldn't it be much easier for DAI to have such transformations done on-host?)

depthai_nodes/ml/messages/classification.py
```diff
@@ -72,3 +72,21 @@ def scores(self, value: NDArray[np.float32]):
...
         self._scores = value
+
+    @property
+    def top_class(self) -> str:
+        """Returns the most probable class.
+
+        @return: The top class.
+        @rtype: str
+        """
+        return self._classes[0]
```
jkbmrz, line 83: Is the first class always the most probably one? Where do we define this?

kkeroo:
> LGTM. The added features seem useful! I'm wondering, though, if we are not making the messages too complicated for DAI to implement natively (e.g. for bbox-related methods, wouldn't it be much easier for DAI to have such transformations done on-host?)

Good question but IIRC Matevz said that we can add functions to msgs with no problem.

depthai_nodes/ml/messages/classification.py
kkeroo: It is sorted in creator function [here](https://github.com/luxonis/depthai-nodes/blob/ef25b049badd86e0d92a679fb0bad2bb4ce65d7e/depthai_nodes/ml/messages/creators/classification.py#L80).

jkbmrz: I see, even though we normally always make the message by using the creator function, I think it's good that message also functions standalone. Thus, I think it's good we add a check to the Scores setter if the values are ordered from the biggest to smallest. Also, I'd write this order requirement in the Classifications message docstring.

kkeroo: The `Classification` msg is also used in ClassificationSequenceParser and `create_classificaiton_sequence_msg` used for text recognition. So, the classes are not ordered by scores there. In this case, using `top_class` wont make sense. Maybe we update the `top_class` docstring with something like `Only works if scores are sorted by descending order`? I could also rename it to `first_class` but IMO `top_class` is better

jkbmrz: Hm true, I'd then go with the docstring explanation as calling the `top_class` method from the `ClassificationSequenceParser` doesn't really make sense (and thus users probably won't make this mistake too often).

PR #125, opened 2024-11-07
Add support for YOLO and SSD in ParserGenerator.

kkeroo:
This PR adds support for YOLO and SSD models in `ParserGenerator`. It uses dai's `DetectionParser` to parse the outputs. We can now instatiate `ParsingNeuralNetwork` node with yolo's or ssd's archives.


klemen1999 (approved):

klemen1999:
Can we also update the examples now? 
And once merged and released we can also update the Model usage doc and utilization scripts since now pipeline won't be different if we use Yolos or other models CC: @jkbmrz

jkbmrz (approved):
LGTM, I'll add a change to the docs!

PR #126, opened 2024-11-08
fix: input size setting

jkbmrz:
Adding a fix to the parsers that require model input size and cannot deduce it from the model outputs (`YuNetParser` and `LaneDetectionParser`) by extending the `build()` method to look for the the `input_shape` within the `head_config`. Additionally, `ParserGenerator` is updated to set the `input_shape` argument aromatically based on the relevant `NN Archive`.

Some additional changes:
- the default `input_size` was removed for the `LaneDetectionParser`,
- all `input_shape` variable/parameter names were changed to `input_size` because it better fits their meaning (shape resembles NCHW/NHWC whereas we only describe HW) → this is more or less a refactoring change only, but it also effects the `input_size` setters for the two parsers (`setInputShape` → `setInputSize`)

The implemented changes were tested on **YuNet** and **Ultra Fast Lane Detection** models.

### DIFF
--- a/depthai_nodes/ml/parsers/lane_detection.py
+++ b/depthai_nodes/ml/parsers/lane_detection.py
@@ -126,23 +126,15 @@ class LaneDetectionParser(BaseParser):
     def build(
         self,
         head_config: Dict[str, Any],
-        inputs_size: List[List[int]],
     ) -> "LaneDetectionParser":
         """Configures the parser.
 
         @param head_config: The head configuration for the parser.
         @type head_config: Dict[str, Any]
-        @param inputs_size: Model inputs size.
-        @type inputs_size: List[List[int]]
         @return: The parser object with the head configuration set.
         @rtype: LaneDetectionParser
         """
 
-        if len(inputs_size) != 1:
-            raise ValueError(
-                f"Only one input supported for LaneDetectionParser, got {len(inputs_size)} inputs."
-            )
-        self.input_size = inputs_size[0]
         output_layers = head_config.get("outputs", [])
         if len(output_layers) != 1:
             raise ValueError(
@@ -154,6 +146,20 @@ class LaneDetectionParser(BaseParser):
         self.cls_num_per_lane = head_config.get(
             "cls_num_per_lane", self.cls_num_per_lane
         )
+
+        inputs = head_config["model"]["inputs"]
+        if len(inputs) != 1:
+            raise ValueError(
+                f"Only one input supported for LaneDetectionParser, got {len(inputs)} inputs."
+            )
+        self.input_shape = inputs[0].get("shape")
+        self.layout = inputs[0].get("layout")
+        if self.layout == "NHWC":
+            self.input_size = [self.input_shape[2], self.input_shape[1]]
+        elif self.layout == "NCHW":
+            self.input_size = [self.input_shape[3], self.input_shape[2]]
+        else:
+            raise ValueError(f"Input layout {self.layout} not supported for input_size extraction.")
 
         return self
 
--- a/depthai_nodes/ml/parsers/lane_detection.py
+++ b/depthai_nodes/ml/parsers/lane_detection.py
@@ -147,7 +147,7 @@ class LaneDetectionParser(BaseParser):
             "cls_num_per_lane", self.cls_num_per_lane
         )
 
-        inputs = head_config["model"]["inputs"]
+        inputs = head_config["model_inputs"]
         if len(inputs) != 1:
             raise ValueError(
                 f"Only one input supported for LaneDetectionParser, got {len(inputs)} inputs."
--- a/depthai_nodes/ml/parsers/lane_detection.py
+++ b/depthai_nodes/ml/parsers/lane_detection.py
@@ -24,7 +24,7 @@ class LaneDetectionParser(BaseParser):
     cls_num_per_lane : int
         Number of points per lane.
     input_shape : Tuple[int, int]
-        Input shape.
+        Input shape (height,width).
 
     Output Message/s
     ----------------
@@ -44,7 +44,7 @@ class LaneDetectionParser(BaseParser):
         row_anchors: List[int] = None,
         griding_num: int = None,
         cls_num_per_lane: int = None,
-        input_shape: Tuple[int, int] = (288, 800),
+        input_shape: Tuple[int, int] = None,
     ) -> None:
         """Initializes the lane detection parser node.
 
@@ -56,7 +56,7 @@ class LaneDetectionParser(BaseParser):
         @type griding_num: int
         @param cls_num_per_lane: Number of points per lane.
         @type cls_num_per_lane: int
-        @param input_shape: Input shape.
+        @param input_shape: Input shape (height,width).
         @type input_shape: Tuple[int, int]
         """
         super().__init__()
@@ -112,7 +112,7 @@ class LaneDetectionParser(BaseParser):
     def setInputShape(self, input_shape: Tuple[int, int]) -> None:
         """Set the input shape for the lane detection model.
 
-        @param input_shape: Input shape.
+        @param input_shape: Input shape (height,width).
         @type input_shape: Tuple[int, int]
         """
         if not isinstance(input_shape, tuple):
--- a/depthai_nodes/ml/parsers/lane_detection.py
+++ b/depthai_nodes/ml/parsers/lane_detection.py
@@ -155,11 +155,13 @@ class LaneDetectionParser(BaseParser):
         self.input_shape = inputs[0].get("shape")
         self.layout = inputs[0].get("layout")
         if self.layout == "NHWC":
-            self.input_size = [self.input_shape[2], self.input_shape[1]]
+            self.input_size = (self.input_shape[2], self.input_shape[1])
         elif self.layout == "NCHW":
-            self.input_size = [self.input_shape[3], self.input_shape[2]]
+            self.input_size = (self.input_shape[3], self.input_shape[2])
         else:
-            raise ValueError(f"Input layout {self.layout} not supported for input_size extraction.")
+            raise ValueError(
+                f"Input layout {self.layout} not supported for input_size extraction."
+            )
 
         return self
 
--- a/depthai_nodes/ml/parsers/lane_detection.py
+++ b/depthai_nodes/ml/parsers/lane_detection.py
@@ -24,7 +24,7 @@ class LaneDetectionParser(BaseParser):
     cls_num_per_lane : int
         Number of points per lane.
     input_shape : Tuple[int, int]
-        Input shape (height,width).
+        Input shape (width,height).
 
     Output Message/s
     ----------------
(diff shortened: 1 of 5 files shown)


depthai_nodes/ml/parsers/yunet.py
```diff
@@ -120,22 +120,26 @@ def setOutputLayerIou(self, iou_output_layer_name: str) -> None:
...
         """
 
         super().build(head_config)
         output_layers = head_config.get("outputs", [])
-        self.input_size = head_config.get("input_size", self.input_size)
-        if len(output_layers) != 3:
+        if len(inputs_size) != 1:
             raise ValueError(
-                f"YuNetParser expects exactly 3 output layers, got {output_layers} layers."
+                f"Only one input supported for LaneDetectionParser, got {len(inputs_size)} inputs."
```
klemen1999, line 139: This is talking about LaneDetectionParser but is inside YuNet?

depthai_nodes/ml/parsers/yunet.py
```diff
@@ -120,22 +120,26 @@ def setOutputLayerIou(self, iou_output_layer_name: str) -> None:
...
             )
+        self.input_size = inputs_size[0]
+        print(self.input_size)
```
klemen1999, line 142: Please remove

depthai_nodes/parser_generator.py
```diff
@@ -53,7 +54,23 @@ def build(self, nn_archive: dai.NNArchive, head_index: int = None) -> Dict:
                 )
 
             head = decode_head(head)
-            parsers[index] = pipeline.create(parser).build(head)
+            sig = inspect.signature(parser.build)
+            if "inputs_size" in sig.parameters:
+                inputs_size = []
+                for input in nn_archive.getConfig().model.inputs:
+                    breakpoint()
```
klemen1999, line 61: Probably needs to be removed?

depthai_nodes/ml/parsers/lane_detection.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         row_anchors: List[int] = None,
         griding_num: int = None,
         cls_num_per_lane: int = None,
-        input_shape: Tuple[int, int] = (288, 800),
+        input_size: Tuple[int, int] = None,
```
klemen1999, line 47: We need to be careful as this requires also updates to the public NNArchives on HubAI as it is a metadata parameter name change

klemen1999 (commented):
Generally this method could work but I wonder if there isn't a nicer way to infer input size based on the shape of the output and maybe some known stride information or something similar? We did it with YOLO but might not be applicable for all models so if this is not the case here then you can disregard this comment.
If we go ahead with using `input_size` let's make sure we update this in the whole package accordingly.

jkbmrz:
> Generally this method could work but I wonder if there isn't a nicer way to infer input size based on the shape of the output and maybe some known stride information or something similar?

I found no way of determining input size for `YuNetParser` and `LaneDetection` models (the problem is that we don't know the size ratio of the inputs and thus cannot reconstruct it based on the flattened outputs).

> If we go ahead with using input_size let's make sure we update this in the whole package accordingly.

Let's wait for others to see what's their position on it and I can then proceed with changing it!

depthai_nodes/ml/parsers/lane_detection.py
jkbmrz: The change is not breaking as the `input_shape` is currently not defined based on the `head_metadata`. I think we should actually remove those parameters from the archives!

depthai_nodes/ml/parsers/yunet.py
jkbmrz: Fixed in [7b999ac](https://github.com/luxonis/depthai-nodes/pull/126/commits/7b999ac26952c60925c31d1969a99e49a018173c).

depthai_nodes/ml/parsers/yunet.py
jkbmrz: Fixed in [1ed14ba](https://github.com/luxonis/depthai-nodes/pull/126/commits/1ed14baf5553f595cd0dc1f46359552579674f7b).

depthai_nodes/parser_generator.py
jkbmrz: Fixed in [1ed14ba](https://github.com/luxonis/depthai-nodes/pull/126/commits/1ed14baf5553f595cd0dc1f46359552579674f7b).

depthai_nodes/ml/parsers/lane_detection.py
klemen1999: I see. In that case this was a bit of a bug because if we exported this model with different input shape the automatic ParserGenerator wouldn't work correctly as input shape would not be changed from default right?

depthai_nodes/ml/parsers/lane_detection.py
```diff
@@ -109,32 +109,40 @@ def setClsNumPerLane(self, cls_num_per_lane: int) -> None:
...
+        if len(input_size) != 2:
+            raise ValueError("Input size must be a tuple of two integers.")
+        if not all(isinstance(size, int) for size in input_size):
+            raise ValueError("Input size must be a tuple of integers.")
+        self.input_size = input_size
 
     def build(
         self,
         head_config: Dict[str, Any],
+        inputs_size: List[List[int]],
```
klemen1999, line 129: Hm this breaks the signature of the build function defined in the BaseParser. What if we rather have build in BaseParser defined as:
```python
def build(self, head_config: Dict[str, Any], **kwargs: Dict[str, Any]) -> "BaseParser"
```

And then any additional params get passed are are accesed through kwargs? So for parsers that would need "input_size" we get it in build method as:
```python
self.input_size = kwargs.get("input_size", self.input_size) # use default if not present and default would be None and we error out in run() if it stays None
```
Thoughs?

depthai_nodes/parser_generator.py
```diff
@@ -53,7 +54,22 @@ def build(self, nn_archive: dai.NNArchive, head_index: int = None) -> Dict:
...
+                for input in nn_archive.getConfig().model.inputs:
+                    if input.layout == "NHWC":
+                        _, height, width, _ = input.shape
+                    elif input.layout == "NCHW":
+                        _, _, height, width = input.shape
+                    else:
+                        raise ValueError(
+                            f"Input layout {input.layout} not supported for input_size extraction."
+                        )
+                    inputs_size.append([width, height])
```
klemen1999, line 69: This should be List of Tuples
```suggestion
                    inputs_size.append(tuple([width, height]))
```

depthai_nodes/parser_generator.py
```diff
@@ -53,7 +54,22 @@ def build(self, nn_archive: dai.NNArchive, head_index: int = None) -> Dict:
...
+            sig = inspect.signature(parser.build)
+            if "inputs_size" in sig.parameters:
+                inputs_size = []
+                for input in nn_archive.getConfig().model.inputs:
+                    if input.layout == "NHWC":
+                        _, height, width, _ = input.shape
+                    elif input.layout == "NCHW":
+                        _, _, height, width = input.shape
+                    else:
+                        raise ValueError(
```
kkeroo, line 66: We shouldn't raise an error because some models will have input layout different from NCHW/NHWC. E.g. Gaze estimation model requires 3rd input to be of shape [1,3] (there are also some embedding models and LLM in the future maybe). What we can do is to set the `input_size` to the whole input shape (e.g. [1,3]) and print a warning saying this input size is used for input `<x>`. Thoughts?

depthai_nodes/ml/parsers/lane_detection.py
kkeroo: This could work.
Alternative option: we extend `head_config` with inputs information and then move the input_size extraction logic from parser_generator to base_parser `build` method. this way we dont introduce new parameters to any functions. But we need to call `super.build()` in all parsers build methods.

depthai_nodes/parser_generator.py
klemen1999: Good point. I agree with passing the whole input_shape and then the parser is responsible for getting input size (height, width) for it since specific parser "knows" what type of inputs it expects (e.g. YuNet expects image input so we can get height and width inside its build()). In case layout is NHWC would it make sense to convert to NCHW here so we always work with NCHW internally?

depthai_nodes/ml/parsers/lane_detection.py
klemen1999: Actually this seems a nicer way of doing it yeah. One note: This makes build() not abstract anymore - is there any concern because of this? I can't think of any but it is something to be aware of

depthai_nodes/ml/parsers/lane_detection.py
jkbmrz: Yes, I think that's right. The `input_shape` was always (288, 800) by default unless changed with the `setInputShape` setter.

depthai_nodes/ml/parsers/lane_detection.py
jkbmrz: Both options are fine to me! Also, I don't see a reason why we would want to keep the parent build() method abstract.

depthai_nodes/parser_generator.py
jkbmrz: > We shouldn't raise an error because some models will have input layout different from NCHW/NHWC.

We only need to set the `inputs_size` parameter for two models for now (YuNet and LineDetection). So this part of the code wouldn't be run for GazeEstimation. However, I see how this is a bit buggy and can cause trouble later on. Will move this part to parser as suggested by @klemen1999.

depthai_nodes/parser_generator.py
jkbmrz: The `inputs_size` parameter is currently defined as `List[List[int]]`. So this would also require a change there. Do we prefer a list of Tuples instead of a list of Lists?

depthai_nodes/parser_generator.py
klemen1999: Yes, let's put this into build() and set `self.input_shape` automatically. And migh be good to check if layout is NHWC to transpose it to NCHW so it is easier to process internally

depthai_nodes/parser_generator.py
```diff
@@ -52,8 +53,13 @@ def build(self, nn_archive: dai.NNArchive, head_index: int = None) -> Dict:
                     f"Parser {parser_name} does not inherit from BaseParser class."
                 )
 
-            head = decode_head(head)
-            parsers[index] = pipeline.create(parser).build(head)
+            head_config = decode_head(head)
+            head_config["model"] = {"inputs": []}
```
aljazkonec1, line 57: I would not do a nested dictionary just for inputs. The entire point of decode_head is to get a dictionary of key-value pairs that has zero nestings. Rather just add `head_config["inputs"] = []` and then do the rest.

jkbmrz: Added in [c781241](https://github.com/luxonis/depthai-nodes/pull/126/commits/c781241884b3774880f93e11dc75d0f01dd48478).

aljazkonec1 (approved):

klemen1999 (approved):

PR #127, opened 2024-11-08
refactor: segmentation messaging

jkbmrz:
Refactoring segmentation messages and creators:
- removed `SegmentationMasksSAM` message (switching to `SegmentationMasks` message instead),
- removed `create_sam_message` creator (switching to `create_segmentation_message` creator instead),
- adjusted the `FastSAMParser` and `SegmentationParser` to the changes,
- changed `SegmentationMasks` message's mask type from `int8` to `int16`.

The code-updates were tested on the following models:
- FastSAM,
- MediaPipe Selfie Segmentation,
- DeepLab-V3-Plus,
- PP-LiteSeg,
- eWaSR.


klemen1999:
> what about if two masks are present at same pixel? we loose this information with the current way!

You are correct but this decision was made because having one seg map for N instances is much more memory efficient than having N separate 2D masks. It is a trade-off between what is possible to have on device and what would be nice to have.

> should we really want to use -1 for background?

What if we rather make mask in `SegmentationMask` an uint8 type (instead of int8) and then have background be 255 (max uint8 value). This is then also more optimized for on-device porting with DAI. IMO limitation of max 255 possible instances seems reasonable? Otherwise we can move to uint16 and have 65535 max instances which should for sure be enough.

> what to do with the _fastsam_show_masks visualizer?

Is this needed? Quickly looking at it is seems like it is just looking at each pixel and merging masks into one by taking the class with highest index. Am I missing something? And there are some morphology operations before - are those part of how original repo preprocessed them and are they needed?

examples/visualization/visualizers/segmentation.py
```diff
@@ -2,14 +2,14 @@
...
+from depthai_nodes.ml.messages import SegmentationMask
 
 from .utils.colors import get_adas_colors, get_ewasr_colors, get_selfie_colors
-from .utils.message_parsers import parse_fast_sam_message, parse_segmentation_message
+from .utils.message_parsers import parse_segmentation_message
 
 
 def visualize_segmentation(
-    frame: dai.ImgFrame, message: dai.ImgFrame, extraParams: dict
+    frame: dai.ImgFrame, message: SegmentationMask, extraParams: dict
```
klemen1999, line 12: I believe frame is actually np.ndarray type here and not ImgFrame. And same for all other `visualize_...` functions

jkbmrz: fixed in [31c4e62](https://github.com/luxonis/depthai-nodes/pull/127/commits/31c4e62bbfc7fb8aca0a41152619a3f16f3dbb94).

jkbmrz:
@klemen1999 1 and 2 are resolved. Regarding 3, is this maybe how we should approach joining masks into one by taking the classes with the highest index? Currently we merge them just by iterating through them and overwriting the existing info.

depthai_nodes/ml/messages/segmentation.py
```diff
@@ -5,26 +5,25 @@
 
 class SegmentationMask(dai.Buffer):
     """SegmentationMask class for a single- or multi-object segmentation mask.
-    Background is represented with "-1" and foreground classes with non-negative
-    integers.
+    Background is represented with "0" and foreground classes with positive integers.
```
klemen1999, line 8: Let's keep actual object classes as they were, starting with 0 and let's make background be the maximum datatype value. So if mask: NDArray[np.uin16] then background is `65535`. You can get this value with:
`np.iinfo(np.uint16).max.`
This way values in the mask correspond 1:1 with instance detections and we also don't use negative number which is more device efficient. Thoughts?

examples/visualization/visualizers/keypoints.py
```diff
@@ -1,12 +1,12 @@
...
-import depthai as dai
+import numpy as np
 
 from depthai_nodes.ml.messages import Keypoints
 
 from .utils.message_parsers import parse_keypoints_message
 
 
-def visualize_keypoints(frame: dai.ImgFrame, message: Keypoints, extraParams: dict):
+def visualize_keypoints(frame: np.ndarray, message: Keypoints, extraParams: dict):
```
klemen1999, line 9: There are other places where this type is incorrectly set. I belieeve this is the case for every `visualize...` function in the `examples/`. Let's correct the types for all of them in this PR.

klemen1999:
> Regarding 3, is this maybe how we should approach joining masks into one by taking the classes with the highest index? 

We can unify this yeah. Although logically the if have a pixel that can be class 0 or class 10 then we just have to pick one, there isn't really anything that tells us which one is more important. But I agree, we should always follow the same convetion and be clear about it in the documentation.

depthai_nodes/ml/messages/segmentation.py
kkeroo: Is there a reason we dont go with 0 as the background class?

depthai_nodes/ml/messages/creators/segmentation.py
```diff
@@ -1,61 +1,33 @@
...
-    @type x: np.array
-    @return: Output segmentaion message in ImgFrame.Type.RAW8.
-    @rtype: dai.ImgFrame
-    @raise ValueError: If the input is not a numpy array.
-    @raise ValueError: If the input is not 3D.
-    @raise ValueError: If the input 3rd dimension is not E{1}.
+    @param mask: Segmentation map array of shape (H, W) where each value represents a
+        segmented object class.
+    @type mask: np.array
+    @return: Segmentaion mask message.
```
kkeroo, line 12: missing `t`

kkeroo (approved):
LGTM overall. Jakob, maybe you also check README in messages where Segmentation message is describe - we should have info for background class and desired datatype (uint8,...)

depthai_nodes/ml/messages/segmentation.py
klemen1999: Yeah, main reason why we don't want this is the case for instance segmentation. There we have list of detections and a joint mask for all of them. And ideally we want the object at index 0 have it's instance mask encoded with index 0.

kkeroo: Okay, thanks for clarification!

depthai_nodes/ml/messages/creators/segmentation.py
jkbmrz: Fixed in [f4394fd](https://github.com/luxonis/depthai-nodes/pull/127/commits/f4394fd488f0bb1ddb7c87808d4a35ee25dac88a).

examples/visualization/visualizers/keypoints.py
jkbmrz: Fixed in [439f4f0](https://github.com/luxonis/depthai-nodes/pull/127/commits/439f4f0cd768189f643bd5cdf7c05f8cf981adbb).

depthai_nodes/ml/messages/segmentation.py
jkbmrz: Added in [8b6db71](https://github.com/luxonis/depthai-nodes/pull/127/commits/8b6db71f66f0dae8cde8be2f9527f90501711da8).

jkbmrz:
I've switched mask types to `int16`. cc @klemen1999 for re-review.

depthai_nodes/ml/parsers/segmentation.py
```diff
@@ -143,10 +143,11 @@ def run(self):
 
             class_map = (
                 np_function(segmentation_mask, axis=0)
-                .reshape(segmentation_mask.shape[1], segmentation_mask.shape[2], 1)
-                .astype(np.uint8)
+                .reshape(segmentation_mask.shape[1], segmentation_mask.shape[2])
+                .astype(np.int16)
             )
+            class_map = class_map - 1
```
klemen1999, line 149: I think this is needed only for the case where `segmentation_mask.shape[0]==1` and we manually add a background class. Or even better what if we instead of np.zeros() there stack an array of -1s? Because for other cases we can't assume that class at "slice" 0 is background class. It could be that we have panoptic segmentation where every pixel is segmented into a distinct class and this might not be "background".

klemen1999 (approved):

PR #130, opened 2024-11-13
Improvement/img detection extended

aljazkonec1:
This PR adapts ImgDetectionsExtended to use `dai.RotatedRect` object for storing information about the (rotated) bounding box. This simplifies porting to depthai in the future and reduces code duplication. Examples were also updated to take advantage of this new definition.

Some smaller updates were also made due to the use of dai v3 alpha6.


klemen1999 (commented):
Generally LGTM, missing getter and setter for label_name. And let's also update README for messages with the new definition of `ImgDetectionExtended`

aljazkonec1:
> Generally LGTM, missing getter and setter for label_name. And let's also update README for messages with the new definition of `ImgDetectionExtended`
I'll update this now and add some other functionality before opening up for review

klemen1999 (approved):

jkbmrz (approved):
LGTM

aljazkonec1:
I see your point @kkeroo but there are two things to consider:
  1)  DAI already has `RotatedRect` and uses it in a lot of functions (like `ImageManipNodeV2`). Therefore, for the user to access these functionalities, he would have to convert to RotatedRect object. Why not do it for them then?
  2) When updating examples (and in some other code), I found the usage of `RotatedRect` to be easier than having to manually write x_center, y_center. I could just do `rect = detection.rotated_rect` and then focus on it as a whole rectangle instead of focusing on every single coordinate/shape on its own. 

I do however still agree on having utils functions for converting to/from "depthai objects" (like `RotatedRect`, `Point2f`, `Size2f`, etc etc) to numpy arrays. But this can be done in another PR where we clearly define these functions and make the syntax as close to depthai syntax as possible.

kkeroo:
> 1. DAI already has RotatedRect and uses it in a lot of functions (like ImageManipNodeV2). Therefore, for the user to access these functionalities, he would have to convert to RotatedRect object. Why not do it for them then?

Valid point. 

> 2. When updating examples (and in some other code), I found the usage of RotatedRect to be easier than having to manually write x_center, y_center. I could just do rect = detection.rotated_rect and then focus on it as a whole rectangle instead of focusing on every single coordinate/shape on its own.

Depends on the visualization. I see you used `cv2.polylines`

```
rect = detection.rotated_rect
points = rect.getPoints()
bbox = np.array([[point.x, point.y] for point in points])
bbox = bbox.astype(int)
cv2.polylines(frame, [bbox], isClosed=True, color=(255, 0, 0), thickness=2)
```

If we just add method to the ImgDetectionExtended we can have it shorter and more intuitive. And inside the method we explicitly `use rect.getPoints`:
```
def get_bbox(self):
    rect = detection.rotated_rect
    points = rect.getPoints()
    bbox = np.array([[point.x, point.y] for point in points])
    bbox = bbox.astype(int)
    return bbox
```

user wil then just call
```
bbox = detection.get_bbox()
cv2.polylines(...)
```

and the same for the rect.getOuterRect(). This just adds the option for getting right bbox coordinates in one line and we keep `RotatedRect` for the cases you provided.

aljazkonec1:
Okay I agree with this style now.

PR #134, opened 2024-11-14
feat: adjust ParsingNeuralNetwork to DAI v3 alpha6

jkbmrz:
Adding an update to ParsingNeuralNetwork build() method to match the recent changes in DAI (v3alpha6) that allow models to be built directly from slug. 

Previous functionality: `pipeline.create(ParsingNeuralNetwork).build(<CameraNode>, <dai.NNModelDescription OR dai.NNArchive>)`

The updated functionality:
`pipeline.create(ParsingNeuralNetwork).build(<CameraNode>, <Slug>)`

### DIFF
--- a/depthai_nodes/parsing_neural_network.py
+++ b/depthai_nodes/parsing_neural_network.py
@@ -39,22 +39,10 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         self._nn = self._pipeline.create(dai.node.NeuralNetwork)
         self._parsers: Dict[int, BaseParser] = {}
 
-    @overload
-    def build(
-        self, input: dai.Node.Output, nn_source: dai.NNModelDescription, fps: int
-    ) -> "ParsingNeuralNetwork":
-        ...
-
-    @overload
-    def build(
-        self, input: dai.Node.Output, nn_source: dai.NNArchive, fps: int
-    ) -> "ParsingNeuralNetwork":
-        ...
-
     def build(
         self,
         input: dai.Node.Output,
-        nn_source: Union[dai.NNModelDescription, dai.NNArchive],
+        slug: str,
         fps: int = None,
     ) -> "ParsingNeuralNetwork":
         """Builds the underlying NeuralNetwork node and creates parser nodes for each
@@ -63,9 +51,8 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         @param input: Node's input. It is a linking point to which the NeuralNetwork is
             linked. It accepts the output of a Camera node.
         @type input: Node.Input
-        @param nn_source: NNModelDescription object containing the HubAI model
-            descriptors, or NNArchive object of the model.
-        @type nn_source: Union[dai.NNModelDescription, dai.NNArchive]
+        @param slug: HubAI model slug.
+        @type slug: str
         @param fps: FPS limit for the model runtime.
         @type fps: int
         @return: Returns the ParsingNeuralNetwork object.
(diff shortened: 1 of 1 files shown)


depthai_nodes/parsing_neural_network.py
```diff
@@ -74,21 +61,25 @@ def build(
...
+                raise ValueError(
+                    "Slug must be in the format <model_slug>:<model_version_slug>."
                 )
-            self._nn_archive = dai.NNArchive(dai.getModelFromZoo(nn_source))
-        elif isinstance(nn_source, dai.NNArchive):
-            self._nn_archive = nn_source
-        else:
-            raise ValueError(
-                "nn_source must be either a NNModelDescription or NNArchive"
+            model_slug, model_version_slug = slug.split(":")
```
aljazkonec1, line 69: Have you tested this? Model slugs can now include ReferenceSlug like: `luxonis/yolov6-nano:r2-coco-512x288:b45ea75`

jkbmrz: We could evaluate this further but I just wanted to make sure that ":" is present so that next line doesn't fail. If the slug is incorrect, the code fails and the error is quite descriptive (`No model found for the given model description`) so I don't think additional checks are needed at this level.

aljazkonec1: Yes but if two ":" are present in a string, the split will produce a list of 3 substrings, not the two defined and the code will throw an error even if valid slug is given.

jkbmrz: Thanks! I've added a fix in [5ff83dd](https://github.com/luxonis/depthai-nodes/pull/134/commits/5ff83ddf96cf68da4e4615f33aa2e6504bdb81ab).

depthai_nodes/parsing_neural_network.py
```diff
@@ -39,22 +39,10 @@ def __init__(self, *args, **kwargs) -> None:
...
-    def build(
-        self, input: dai.Node.Output, nn_source: dai.NNArchive, fps: int
-    ) -> "ParsingNeuralNetwork":
-        ...
-
     def build(
         self,
         input: dai.Node.Output,
-        nn_source: Union[dai.NNModelDescription, dai.NNArchive],
+        model: str,
```
aljazkonec1, line 45: I was thinking about this, and remembered we said to keep the same syntax as `dai.NeuralNetwork` node, currently there are 2 accepted build functions: `.build( input: Camera, modelDesc: dai.NNModelDescription, fps: float)`and `.build(input: Camera, nnArchive: dai.NNArchive, fps: float)`. For keeping in sync with depthai syntax/function definitions, I would split the this build method in the same way.

aljazkonec1 (commented):
Looks good, just one comment about keeping up with depthai definitions

depthai_nodes/parsing_neural_network.py
jkbmrz: cc @klemen1999 on this. We can add overrides and allow the old `nn_source` types. But the option of using slug should stay as it comes quite handy (e.g. when re-writing experiments).

klemen1999: Can we keep all three options? Basically add that nn_source can also be str and handle this properly inside the build()?

jkbmrz: Adding in [5d2208f](https://github.com/luxonis/depthai-nodes/pull/134/commits/5d2208faf8a891d59bea2a82b39db51352f2b69f).

examples/main.py
```diff
@@ -31,20 +29,18 @@
...
+    if "gray" in model:
+        image_type = dai.ImgFrame.Type.GRAY8
+
     # YOLO and MobileNet-SSD have native parsers in DAI - no need to create a separate parser
     if parser_name == "YOLO" or parser_name == "SSD":
-        network = pipeline.create(dai.node.DetectionNetwork).build(
-            cam.requestOutput(
-                input_shape, type=dai.ImgFrame.Type.BGR888p, fps=fps_limit
-            ),
+        nn = pipeline.create(dai.node.DetectionNetwork).build(
```
klemen1999, line 38: Note: ParsingNeuralNetwork can now be also used for models with "YOLO and "SSD" parsers so no need for separate if here anymore. But this can be part of separate PR.

klemen1999 (approved):
LGTM

kkeroo (approved):

klemen1999 (approved):

PR #135, opened 2024-11-18
Force NCHW tensor order.

kkeroo:
This PR forces NCHW tensor output when calling `getTensor(...)` method instead of manually checking the shapes and transposing them inside parsers.

Affected Parsers:
- MLSD parser
- FastSAM parser
- YOLOExtended parser
- SuperAnimal parser: removed unnecessary check for shapes.

One yolo util function (`reshape_seg_outputs`) is removed - not needed anymore.

Additionally, `ImgDetectionExtended`'s `masks` parameter now expects int16 and not int8 (introduced in some previous PR) + detection creator function converts `masks` to int16 instead of int8.

### DIFF
--- a/depthai_nodes/ml/messages/creators/detection.py
+++ b/depthai_nodes/ml/messages/creators/detection.py
@@ -160,8 +160,8 @@ def create_detection_message(
         if len(masks.shape) != 2:
             raise ValueError(f"Masks should be of shape (H, W), got {masks.shape}.")
 
-        if masks.dtype != np.int8:
-            masks = masks.astype(np.int8)
+        if masks.dtype != np.int16:
+            masks = masks.astype(np.int16)
 
     detections = []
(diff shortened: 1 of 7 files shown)


jkbmrz (commented):
Looks good but why are we not adding this feature to all parsers?

kkeroo:
Other parsers dont explicitly check for the tensor order - it doesnt matter. E.g.:
- Classification parser: output is not NCHW and it doesnt make sense to be
- Keypoints parser: automatically reshapes the output based on `num_kpts`

and similar for other not-affected parsers. I think forcing `storageOrder=NCHW` in every `getTensor` method is not needed and not ideal because DAI could fail transforming it if the shapes are weird

jkbmrz (approved):

klemen1999 (approved):

PR #140, opened 2024-11-26
E2E Tests: ImageManip node and better checks.

kkeroo:
Checks added in E2E tests:
-  Catches the error when downloading the model from ZOO
-  Catches the error when loading the NN archive
-  Checks if the model has more than one output

Every catch has a `print(e)` so we know whats wrong when running it with `manual.py`

This PR also adds `ImageManipV2` node to the pipeline if the model input size is not divisible by 2 (e.g. 513x513 in deeplab) or the input size is too small. We request 2 or 4 times bigger input size from the camera and then resize it down to model's input size with `ImageManipV2` node.

This feature now runs more models in E2E testing.

### DIFF
--- a/tests/end_to_end/manual.py
+++ b/tests/end_to_end/manual.py
@@ -1,7 +1,7 @@
 import argparse
 
 import depthai as dai
-from utils import get_input_shape, parse_model_slug
+from utils import get_input_shape, get_num_inputs, parse_model_slug
 
 from depthai_nodes.parsing_neural_network import ParsingNeuralNetwork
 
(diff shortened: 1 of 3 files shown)


klemen1999 (approved):

kkeroo:
@klemen1999 I am re-requesting your review since I updated the PR with the addition of `ImageManip` to handle small input sizes.

tests/end_to_end/manual.py
```diff
@@ -81,19 +81,32 @@
         device.close()
         exit(8)
 
+    image_type = (
+        dai.ImgFrame.Type.BGR888i
+        if device.getPlatform().name == "RVC4"
+        else dai.ImgFrame.Type.BGR888p
+    )
```
klemen1999, line 88: Can we set the type based on ˙dai_type˙ field in the NNArchive if present otherwise use this as default?

klemen1999 (commented):
LGTM, just left one suggestion

jkbmrz (approved):

tests/end_to_end/manual.py
kkeroo: Will be addressed in the future because dai has not yet implemented it.

PR #141, opened 2024-11-26
Add visualization methods to all message types

aljazkonec1:
This PR adds default `getVisualizationMessage` methods to all messages. These methods allow depthai to automatically annotate outputs of all models that use these message types. This is the "default" visualizations, if a user wants custom or different annotations, he would have to override these functions or create a new AnnotationNode.

The annotations were tested using RVC4 and RVC2 (where aplicable) on the following models:
YuNet, YOLOv6, YOLOv8, text detection, M-LSD, Fast SAM, Selfie segmentation, PP-LiteSeg, FaceLandmarker, OCR, Emotion classification, Depthanything, Midas, ESRGAN, Ultra Fast Lane Detection


aljazkonec1:
In Keypoints message there is a bug with Visualizer, where it does not show points, for the time being I changed to lines instead.

depthai_nodes/ml/messages/creators/image.py
```diff
@@ -51,3 +51,14 @@ def create_image_message(
     imgFrame.setType(img_frame_type)
 
     return imgFrame
+
+
+def getVisualizationMessage(self) -> dai.ImgFrame:
+    img_frame = dai.ImgFrame()
+    mask = self._map.copy()
```
klemen1999, line 58: What is the `self` in this case? This function looks like a copy of the one in `Map2D` type.

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -226,3 +233,49 @@ def masks(self, value: NDArray[np.int16]):
...
+                ]
+                keypointAnnotation = dai.PointsAnnotation()
+                keypointAnnotation.type = (
+                    dai.PointsAnnotationType.LINE_STRIP
+                )  # change to POINTS when fixed
+                keypointAnnotation.points = dai.VectorPoint2f(keypoints)
+                keypointAnnotation.outlineColor = KEYPOINT_COLOR
+                keypointAnnotation.thickness = 2.0
+                annotation.points.append(keypointAnnotation)
+
```
klemen1999, line 278: Do we also need to add instance seg mask here?

klemen1999 (commented):
Generally LGTM, added some clarifying questions.

depthai_nodes/ml/messages/img_detections.py
aljazkonec1: Sadly we cannot do this. The getVisualizationMessage can only return dai.ImgAnnotations or dai.ImgFrame, and not both at the same time. To show also seg masks, we need to make a new HostNode where we return the seg mask.

depthai_nodes/ml/messages/creators/image.py
aljazkonec1: I must have copied it from Map2D by mistake. It's removed now.

depthai_nodes/ml/messages/map.py
```diff
@@ -71,3 +72,16 @@ def height(self) -> int:
         @rtype: int
         """
         return self._height
+
+    def getVisualizationMessage(self) -> dai.ImgFrame:
+        """Returns default visualization message for 2D maps in the form of a
+        colormapped image."""
+        img_frame = dai.ImgFrame()
+        mask = self._map.copy()
+        if np.any(mask < 1):
```
jkbmrz, line 81: I think it would be safer to do min-max scaling here:
```
min_val = map.min()
    max_val = map.max()
    if map.min() != map.max():
        map = (map - min_val) / (max_val - min_val)
```
This way we always visualize a relative map.

depthai_nodes/ml/messages/segmentation.py
```diff
@@ -48,3 +49,29 @@ def mask(self, value: NDArray[np.int16]):
...
+        unique_values = np.unique(mask[mask >= 0])
+        scaled_mask = np.zeros_like(mask, dtype=np.uint8)
+
+        if unique_values.size == 0:
+            return img_frame.setCvFrame(scaled_mask, dai.ImgFrame.Type.BGR888i)
+
+        min_val, max_val = unique_values.min(), unique_values.max()
+
+        if min_val == max_val:
+            scaled_mask = np.ones_like(mask, dtype=np.uint8) * 255
```
jkbmrz, line 67: I don't get this. min_val == max_val means there is one class present on top of background. Why do we then set all mask values to 255, ignoring the background?

depthai_nodes/ml/messages/segmentation.py
```diff
@@ -48,3 +49,29 @@ def mask(self, value: NDArray[np.int16]):
...
+        else:
+            scaled_mask = ((mask - min_val) / (max_val - min_val) * 255).astype(
+                np.uint8
+            )
+        scaled_mask[mask == -1] = 0
```
jkbmrz, line 72: The background class becomes the same as the first class?

jkbmrz (commented):
Generally LGTM. Adding a few comments.

depthai_nodes/ml/messages/classification.py
```diff
@@ -91,3 +96,28 @@ def top_score(self) -> float:
...
+        """Returns default visualization message for classification.
+
+        The message adds the top five classes and their scores to the right side of the
+        image.
+        """
+        img_annotations = dai.ImgAnnotations()
+        annotation = dai.ImgAnnotation()
+
+        soritng_indexes = np.argsort(self._scores)[::-1]
+        soritng_indexes = soritng_indexes[:5]
```
kkeroo, line 110: Isnt the sorting already handled in the creation function?

kkeroo (approved):
LGTM, 2 small comments/questions

depthai_nodes/ml/messages/classification.py
aljazkonec1: yes but just in case I wanted to do it again if the Classfication message is used in some other parser/message_creator

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -226,3 +233,49 @@ def masks(self, value: NDArray[np.int16]):
...
+            pointsAnnotation.type = dai.PointsAnnotationType.LINE_STRIP
+            pointsAnnotation.points = dai.VectorPoint2f(points)
+            pointsAnnotation.outlineColor = OUTLINE_COLOR
+            pointsAnnotation.thickness = 2.0
+            annotation.points.append(pointsAnnotation)
+
+            text = dai.TextAnnotation()
+            text.position = points[0]
+            text.text = f"{detection.label_name} {int(detection.confidence * 100)}%"
+            text.fontSize = 50.5
```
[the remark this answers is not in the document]
aljazkonec1: I just kept it the same as it is in visualizer example. I havent found a way to dynamically set the text size sadly.

depthai_nodes/ml/messages/segmentation.py
aljazkonec1: I could also set it to any other value, just personal preference for 255 and then color map 255 to the correct color

depthai_nodes/ml/messages/segmentation.py
aljazkonec1: yes because there cant be negative values in the mask when doing color mapping. Later, when the colormapping is done, the colors for background are set to [0, 0, 0] in line 74: `colored_mask[mask == -1] = [0, 0, 0]`

depthai_nodes/ml/messages/segmentation.py
jkbmrz: oh I see, my bad, its resolved below with` colored_mask[mask == -1] = [0, 0, 0]`

klemen1999 (approved):

PR #142, opened 2024-11-28
Add dai.ImgTransformations to messages and improvements to Text detection parser

aljazkonec1:
This PR adds the following:
 1) `dai.ImgTransformations` attribute to all depthai-nodes messages. This is necessary as the information inside dai.ImgTransofrmations is needed in later nodes for dai.ImgManip, and for visualization purposes as width, height and other values are stored in this attribute.
2) adds some improvements to Text Detection parser and Palm detection parser where the angle is rounded to a full degree.
3) Make Text detection parser more robust and simplify some postprocessing functions, making the detections much more reliable

### DIFF
--- a/depthai_nodes/ml/messages/classification.py
+++ b/depthai_nodes/ml/messages/classification.py
@@ -14,6 +14,8 @@ class Classifications(dai.Buffer):
         A list of classes.
     scores : NDArray[np.float32]
         Corresponding probability scores.
+    transformation : dai.ImgTransformation
+        Image transformation object.
     """
 
     def __init__(self):
@@ -21,6 +23,7 @@ class Classifications(dai.Buffer):
         dai.Buffer.__init__(self)
         self._classes: List[str] = []
         self._scores: NDArray[np.float32] = np.array([])
+        self._transformation: dai.ImgTransformation = None
 
     @property
     def classes(self) -> List:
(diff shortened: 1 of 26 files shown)


depthai_nodes/ml/parsers/mediapipe_palm_detection.py
```diff
@@ -192,9 +192,12 @@ def run(self):
 
             bboxes = np.clip(bboxes, 0, 1)
             points = np.clip(points, 0, 1)
+            angles = np.round(angles, 0)
 
             detections_msg = create_detection_message(
-                bboxes=bboxes, scores=scores, angles=angles, keypoints=points
+                bboxes=bboxes, scores=scores, angles=angles
```
klemen1999, line 198: is removing `keypoints=points` wanted here?

klemen1999 (approved):
LGTM

depthai_nodes/ml/messages/classification.py
```diff
@@ -91,3 +94,26 @@ def top_score(self) -> float:
         @rtype: float
         """
         return self._scores[0]
+
+    @property
+    def transformation(self) -> dai.ImgTransformation:
+        """Returns the Image Transformation object.
```
jkbmrz, line 100: As the transformation property and setter are the same for every message, we could also make a message parent class that would define them once instead of copy-pasting them to all messages.

jkbmrz (approved):
Added a small remark but LGTM otherwise!

kkeroo (approved):

depthai_nodes/ml/parsers/mediapipe_palm_detection.py
aljazkonec1: Mistake, I was testing something and removed it.

depthai_nodes/ml/messages/classification.py
aljazkonec1: Yes I also thought about it. I asked depthai team and when this will get ported, all messages will inherit from `dai.Buffer` and will have similar structure as `dai.ImgDetections` currently has. Thats why I left it as is.

PR #143, opened 2024-12-02
Integration tests for parsers.

kkeroo:
This PR adds integration tests for parsers. Tests work in a similar way as end-2-end tests. We have `main.py` where we can locally run a test and we can trigger the tests in the GH actions.

These tests are designed to simulate a neural network output and send it to the parser node. The parser node will parse the output and return the message. The tests check if the parser outputs the expected message. Since we know what `NNData` we are sending to the parser, we can check if the outputted message is correct.

The testing pipeline is as follows:

1. Load the neural network archive.
2. Load the parser node with `ParserGenerator` based on the NN archive.
3. Read and load the `NNData` from pickle file. It contains the output of the neural network.
4. Send the `NNData` to the parser node.
5. Read and load the expected message from the pickle file.
6. Check if the parser node outputs the expected message.

## Test generation

To generate a new test for the parser, you can use `extract_nn_data.py` script. The script will extract the `NNData` from the neural network output and store it in the pickle file. The script requires the following arguments: `-m` for the model slug, `-img` for the input image, and optional `-ip` for the device IP or mxid.

The script does not generate the expected message because each parser has its own message format and DAI messages can not be dumped in the pickle file.

One example for generating the expected message for the `ClassificationParser`:

```python
expected_output = {
    'model': 'luxonis/efficientnet-lite:lite0-224x224',
    'parser': 'ClassificationParser',
    'classes': ['wolf', 'dog', 'cat'],
    'scores': [0.7, 0.2, 0.1]
}

with open('nn_datas/ClassificationParser/efficientnet-lite_output.pkl', 'wb') as f:
    pickle.dump(expected_output, f)
```

## Running the tests

To run the tests, you can use the `main.py` script. You can use `--all` flag to test all parsers or test a specific parser with `-p` flag.
You would need the B2 credentials to download the tests from the bucket and set it in the ENV variables `B2_APPLICATION_KEY_ID` and `B2_APPLICATION_KEY`.

## Whats left?
This is the inital version of integration tests. All exisiting parsers have a tests prepared except the Regression and XFeat parsers. I had some troubles with them. Will add it shortly but I want to merge this PR so we can run the tests in the CI.

## Why storing NNDatas in file?
I think this is the easiest way of doing it because the framework for loading and generating NNDatas is simple and with new parsers we only need to run `extract_nn_data.py` to save the NNData to a file and then create an expected message manually. Sadly, we can not dump the DAI classes to pickle file. If anyone has better way, please share it:)

P.S.: CI successfully finished in the private repo.

### DIFF
--- a/tests/integration_tests/check_messages.py
+++ b/tests/integration_tests/check_messages.py
@@ -12,6 +12,7 @@ from depthai_nodes.ml.messages import (
     Keypoints,
     Lines,
     Map2D,
+    Predictions,
     SegmentationMask,
 )
 
@@ -278,6 +279,27 @@ def check_line_msg(message: Lines, expected_output: Dict[str, Any]):
         ), f"Expected end_point {expected_line['end_point']}, got {predicted_line['end_point']}."
 
 
+def check_regression_msg(message: Predictions, expected_output: Dict[str, Any]):
+    """
+    Expected output format:
+    {
+        "model": "luxonis/gaze-estimation-adas:60x60:0.0.1",
+        "parser": "RegressionParser",
+        "value": [0.4, 0.2, 0.1]
+    }
+    """
+    predictions = message.predictions
+    predictions = np.array([pred.prediction for pred in predictions])
+    expected_predictions = np.array(expected_output["value"])
+    print(
+        f"Expected predictions shape: {expected_predictions.shape}, predicted predictions shape: {predictions.shape}"
+    )
+    assert (
+        predictions.shape == expected_predictions.shape
+    ), f"The shape of the predictions is different. Expects {expected_predictions.shape}, got {predictions.shape}"
+    np.testing.assert_allclose(predictions, expected_predictions, rtol=1e-2)
+
+
 def test_output(message, model_slug, parser_name):
     expected_output = load_expected_output(model_slug, parser_name)
 
@@ -305,6 +327,8 @@ def test_output(message, model_slug, parser_name):
         check_line_msg(message, expected_output)
     elif expected_output["parser"] == "PPTextDetectionParser":
         check_detection_msg(message, expected_output)
+    elif expected_output["parser"] == "RegressionParser":
+        check_regression_msg(message, expected_output)
     elif expected_output["parser"] == "SCRFDParser":
         check_detection_msg(message, expected_output)
     elif expected_output["parser"] == "SegmentationParser":
--- a/tests/integration_tests/check_messages.py
+++ b/tests/integration_tests/check_messages.py
@@ -33,6 +33,10 @@ def check_classification_msg(message: Classifications, expected_output: Dict[str
         "score": 0.9
     }
     """
+    assert isinstance(
+        message, Classifications
+    ), f"The message is not a Classifications. Got {type(message)}."
+
     print(
         f"Expected top class: {expected_output['class']}, predicted top class: {message.top_class}"
     )
@@ -53,6 +57,10 @@ def check_classification_sequence_msg(
         "parser": "ClassificationSequenceParser",
         "class": ['HELLO']
     """
+    assert isinstance(
+        message, Classifications
+    ), f"The message is not a Classifications. Got {type(message)}."
+
     print(
         f"Expected top class: {expected_output['class']}, predicted top class: {message.classes}"
     )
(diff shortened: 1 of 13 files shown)


tests/integration_tests/extract_nn_data.py
```diff
@@ -0,0 +1,108 @@
...
+import pickle
+
+import cv2
+import depthai as dai
+from utils import extract_main_slug, extract_parser, get_input_shape
+
+argparser = argparse.ArgumentParser()
+argparser.add_argument("-ip", help="IP address of the device", default="")
+argparser.add_argument(
+    "-img", "--img_path", help="Path to the input image", required=True, type=str
```
klemen1999, line 12: Do we have the test images also stored together with the outputs the models generated? It would be good to have them so we can reuse them maybe in the future.

klemen1999 (approved):
Nice LGTM!

kkeroo:
### Update on Regression parser

Test for regression parser has been added in [c99bbc4](https://github.com/luxonis/depthai-nodes/pull/143/commits/c99bbc4e7173167646afee00f4d2202400c53fb3).

tests/integration_tests/extract_nn_data.py
kkeroo: Yes, it is saved along with the outputs in the `extract_nn_data.py` [here](https://github.com/luxonis/depthai-nodes/blob/c99bbc4e7173167646afee00f4d2202400c53fb3/tests/integration_tests/extract_nn_data.py#L104).

aljazkonec1 (approved):

tests/integration_tests/check_messages.py
```diff
@@ -0,0 +1,341 @@
...
+        assert np.allclose(
+            expected_detection["angle"], predicted_detection["angle"], atol=0.01
+        ), f"Expected angle {expected_detection['angle']}, got {predicted_detection['angle']}."
+
+        if "keypoints" in expected_detection.keys():
+            assert np.allclose(
+                expected_detection["keypoints"],
+                predicted_detection["keypoints"],
+                atol=0.01,
+            ), f"Expected keypoints {expected_detection['keypoints']}, got {predicted_detection['keypoints']}."
```
jkbmrz, line 222: This check for keypoints similarity is different than the one used in the `check_keypoints_msg`. Should we do a same check?

jkbmrz (approved):
LGTM! Just adding a few small non-blocking comments.

tests/integration_tests/check_messages.py
```diff
@@ -0,0 +1,341 @@
...
+    print(
+        f"Expected embeddings shape: {expected_embeddings.shape}, predicted embeddings shape: {embeddings.shape}"
+    )
+    assert (
+        embeddings.shape == expected_embeddings.shape
+    ), "The shape of the embeddings is different."
+    np.testing.assert_allclose(embeddings, expected_embeddings, rtol=1e-2)
+
+
+def check_fastsam_msg(
```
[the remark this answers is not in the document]
kkeroo: Thanks for noticing! It should be called `check_segmentation_msg()`. Fixed in [7ae446b](https://github.com/luxonis/depthai-nodes/pull/143/commits/7ae446bfcdcc7a47ea243b43fbd438e63b812dd2).

tests/integration_tests/check_messages.py
```diff
@@ -0,0 +1,341 @@
...
+    total = expected_mask.size
+    acc = correct / total
+    print(f"Accuracy: {acc}")
+    assert (
+        acc > threshold
+    ), f"The accuracy {acc} is lower than the threshold {threshold}"
+
+
+def check_keypoints_msg(
+    message: Keypoints,
```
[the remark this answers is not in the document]
kkeroo: Good idea. Added in [08c08cf](https://github.com/luxonis/depthai-nodes/pull/143/commits/08c08cf1a6dba261918ceb6481f55bc165f20b65).

tests/integration_tests/check_messages.py
```diff
@@ -0,0 +1,341 @@
...
+
+    for gt_kpt in expected_keypoints:
+        min_dist = float("inf")
+        for pred_kpt in keypoints:
+            dist = (gt_kpt[0] - pred_kpt[0]) ** 2 + (gt_kpt[1] - pred_kpt[1]) ** 2
+            if dist < min_dist:
+                min_dist = dist
+        assert (
+            min_dist < distance_threshold
+        ), f"Expected min distance < {distance_threshold}, got {min_dist}"
```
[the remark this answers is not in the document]
jkbmrz: Aha, got it! Then it's all good, missed the part where you go over all the GT keypoints.

tests/integration_tests/check_messages.py
kkeroo: Thanks for noticing, fixed in [13407fa](https://github.com/luxonis/depthai-nodes/pull/143/commits/13407fab7997f86ff2b3921c3c4242bae007e104).

PR #144, opened 2024-12-03
[FIX] dai.ImgTransformations is None for RVC2 device

aljazkonec1:
This PR adds a hotfix for using RVC2 device with depthai v3_develop. The issue is that the dai.NNdata has self.transformation set to `None`. This is a know issue in depthai and will be fixed in comming updates.

The PR was tested on both RVC2 and RVC4 on the following models:
lane detection, depth anything, face landmarks, yolov8 instance seg, selfie seg, scfrd face detection, yolo v6, text detection

### DIFF
--- a/depthai_nodes/ml/messages/classification.py
+++ b/depthai_nodes/ml/messages/classification.py
@@ -112,8 +112,10 @@ class Classifications(dai.Buffer):
         @type value: dai.ImgTransformation
         @raise TypeError: If value is not a dai.ImgTransformation object.
         """
-        if not isinstance(value, dai.ImgTransformation):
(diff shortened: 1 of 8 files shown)


kkeroo:
I suggest that we dont merge features to main if they dont work with alpha releases because they can break the code and parsers dont work. Can we keep features in separate branches till alphas support them? Thoughts?

kkeroo (approved):

aljazkonec1:
Agreed, it was done on alpha6, just that I did not tryout RVC2 afterthe fact.

jkbmrz (approved):

PR #149, opened 2024-12-05
Expose parsers when building a pipeline through ParsingNeuralNetwork

ptoupas:
- Added getParser() method to ParsingNeuralNetwork Class with a default value for `parserID = 0`

### DIFF
--- a/depthai_nodes/parsing_neural_network.py
+++ b/depthai_nodes/parsing_neural_network.py
@@ -129,9 +129,10 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
 
     def getParser(self, parserID: int = 0) -> BaseParser:
         """Returns the parser node for the specified model head."""
-        assert (
-            parserID in self._parsers
-        ), f"Parser with ID {parserID} not found. Available parser IDs: {list(self._parsers.keys())}"
+        if parserID not in self._parsers:
+            raise KeyError(
+                f"Parser with ID {parserID} not found. Available parser IDs: {list(self._parsers.keys())}"
+            )
         return self._parsers[parserID]
 
     def setBackendProperties(self, setBackendProperties: Dict[str, str]) -> None:
--- a/depthai_nodes/parsing_neural_network.py
+++ b/depthai_nodes/parsing_neural_network.py
@@ -127,6 +127,13 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         """Sets the backend of the NeuralNetwork node."""
         self._nn.setBackend(setBackend)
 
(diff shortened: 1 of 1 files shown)


depthai_nodes/parsing_neural_network.py
```diff
@@ -127,6 +127,13 @@ def setBackend(self, setBackend: str) -> None:
         """Sets the backend of the NeuralNetwork node."""
         self._nn.setBackend(setBackend)
 
+    def getParser(self, parserID: int = 0) -> BaseParser:
+        """Returns the parser node for the specified model head."""
+        assert (
```
klemen1999, line 132: Let's go with if check and KeyError because assert can be tehnically ommited if you run the code in optimized mode (-O flag)

depthai_nodes/parsing_neural_network.py
```diff
@@ -127,6 +127,13 @@ def setBackend(self, setBackend: str) -> None:
         """Sets the backend of the NeuralNetwork node."""
         self._nn.setBackend(setBackend)
 
+    def getParser(self, parserID: int = 0) -> BaseParser:
```
klemen1999, line 130: What are your thoughts on having default value here? The other option is that you have to explicitly speciffy the parserID but I agree that in 90% of the cases users are dealing with 1parser NNs so this would be then the desired behaviour. CC: @jkbmrz @kkeroo @aljazkonec1

klemen1999 (approved):
Generally LGTM

depthai_nodes/parsing_neural_network.py
jkbmrz: We could also set the default value to `None` and if no value is provided:
- return the parser at zero index if there is only one parser available,
- error out if there is multiple parsers available to avoid mistakes.

BTW, is `parserID` a good naming? I think `parser_index` would be more appropriate.

aljazkonec1: Having an int value set to None would work but could cause issues down the line when porting to depthai. The function name `getParser` implies that only one is returned and I would just update the docstring such that it explicitly states which one is returned ( i.e. something like: The first parser is returned by default if multiple parsers available).

Regarding the naming, I would switch to snake case (to be consistent with other variables/attributes). `index` does make slightly more sense.

kkeroo: I would leave the default option, so calling `getParser()` returns the first one. I'm fine with both ways of doing it (Petros' way and Jakob's way).

kkeroo (approved):

aljazkonec1 (approved):

PR #156, opened 2025-01-03
Update Point2f and Size2f to use normalized

ptoupas:
Add `normalized` flag into Point2f and Size2f during the creation of `rotated_rect`.
Updated README to point to `depthai==3.0.0a11`


klemen1999 (approved):
LGTM

PR #161, opened 2025-01-22
Tests: Move tests running to self-hosted runners.

kkeroo:
This PR moves E2E and integration tests execution to self-hosted organization runners. This eliminates the need for Wireguard.


kkeroo:
Successful test [here](https://github.com/luxonis/depthai-nodes/actions/runs/12912978165). It is run on oak4-pro because default oak4-s is reserved.

PR #163, opened 2025-01-24
Host nodes: DepthMerger.

kkeroo:
This PR adds first host node called DepthMerger. The node merges normal `ImgDetections` or `ImgDetectionsExtended` with depth information into `SpatialImgDetections`. The node is taken from experiments host-nodes folder and slightly adjusted so it works with `ImgDetectionsExtended`. We also introduce parameter `shrinking_factor` that is responsible for optionally shrinking the ROI for calculating depth values. This is useful because sometimes bounding-box also includes background and so the mean depth value is affected. E.g. In bouding-box of a human the shrinking factor will shrink the ROI into the human torso.

I am not sure about the folder hierarchy. I was thinking of placing it in the `helpers` folder under specific subfolders. Maybe there is something better?

The proposed node works with rewritten experiments: Human-Machine safety and social distancing.


dominik737 (commented):
Looks good. I would have a `nodes` folder along (same level as) the `ml` folder. Since many of the nodes are actually not related to ml. Then if there are too many nodes, we can group then in the `nodes` folder into subfolders.

dominik737:
@klemen1999  Also I just remembered we were talking about tests. Should they be implemented now in those PRs that add nodes?

depthai_nodes/ml/helpers/depth/depth_merger.py
```diff
@@ -0,0 +1,144 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.output = self.createOutput(
+            possibleDatatypes=[
+                dai.Node.DatatypeHierarchy(dai.DatatypeEnum.SpatialImgDetections, True)
+            ]
+        )
+
+        self.shrinking_factor = 0
```
klemen1999, line 36: Make this also configurable through init

depthai_nodes/ml/helpers/depth/host_spatials_calc.py
```diff
@@ -0,0 +1,171 @@
+# HostSpatialsCalc implementation taken from here:
```
klemen1999, line 1: This comment can be removed

depthai_nodes/ml/helpers/depth/host_spatials_calc.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(
+        self,
+        calib_data: dai.CalibrationHandler,
+        depth_alignment_socket: dai.CameraBoardSocket = dai.CameraBoardSocket.CAM_A,
+    ):
+        self.calibData = calib_data
+        self.depth_alignment_socket = depth_alignment_socket
+
+        # Values
+        self.DELTA = 5  # Take 10x10 depth pixels around point for depth averaging
```
klemen1999, line 44: Let's make this 3 values also configurable through init with default values

klemen1999 (commented):
Agree on moving nodes into separate folder (same level as ml) and not nesting them into subfolders at least for now, we can do that later on if needed.

> @klemen1999 Also I just remembered we were talking about tests. Should they be implemented now in those PRs that add nodes?

Good catch, we should make tests for every new node as potentially many different experiments will rely on their specifics. We can sync offline what/how to test.

depthai_nodes/ml/helpers/depth/__init__.py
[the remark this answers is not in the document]
kkeroo: Added in [3738ffd](https://github.com/luxonis/depthai-nodes/pull/163/commits/3738ffd0ff2f551c43e421f43aa0aa016e10fb0d).

depthai_nodes/ml/helpers/depth/depth_merger.py
kkeroo: Added in [3738ffd](https://github.com/luxonis/depthai-nodes/pull/163/commits/3738ffd0ff2f551c43e421f43aa0aa016e10fb0d).

depthai_nodes/ml/helpers/depth/host_spatials_calc.py
kkeroo: Removed in [3738ffd](https://github.com/luxonis/depthai-nodes/pull/163/commits/3738ffd0ff2f551c43e421f43aa0aa016e10fb0d).

depthai_nodes/ml/helpers/depth/host_spatials_calc.py
kkeroo: Added in [3738ffd](https://github.com/luxonis/depthai-nodes/pull/163/commits/3738ffd0ff2f551c43e421f43aa0aa016e10fb0d).

kkeroo:
Okay, I moved the added DepthMerger node to the `depthai_nodes/nodes` folder and exposed it in the `__init__.py`. We can import it as `from depthai_nodes.nodes import DepthMerger`.

depthai_nodes/nodes/depth_merger.py
```diff
@@ -0,0 +1,146 @@
+from typing import Union
+
+import depthai as dai
+
+from depthai_nodes.ml.messages import ImgDetectionExtended, ImgDetectionsExtended
```
jkbmrz, line 5: Keeping consistency with other parts of the library (e.g. parsers), should we make imports from within the library instead? so using `from ..ml.messages import ...` instead of `from depthai_nodes.ml.messages import ...`

jkbmrz (commented):
LGTM. Just adding a small stylistic suggestion.

Also, I agree on having the `nodes` folder at the same level as `ml`. I'm wondering, though, if we should also move there the contents of `depthai_nodes/ml/helpers` folder?

depthai_nodes/nodes/depth_merger.py
klemen1999: In the long run the second (so `from depthai_nodes.ml.messages import ...`) is easier to work with because otherwise there are annoying issues if you for example want to start depth_merger.py as a standalone file - this is atleast my past experience.

jkbmrz: might it then be good we also change the relative imports elsewhere in the library?

klemen1999: +1

jkbmrz: OK, let's add this in another PR (I have some other small changes to add and can include in the same one)

kkeroo:
I added the tests for `DepthMerger` and `HostSpatialsCalc` in [adc4f65](https://github.com/luxonis/depthai-nodes/pull/163/commits/adc4f651783382049ab97d75d95d0d2eab41c64a).

PR #166, opened 2025-01-28
Update ImgExtractedDetection: Soften confidence check

HonzaCuhel:
This PR softens the confidence check in `ImgDetectionExtended` class, so that values like `1.0000001192092896` would pass.

### DIFF
--- a/depthai_nodes/ml/messages/img_detections.py
+++ b/depthai_nodes/ml/messages/img_detections.py
@@ -83,7 +83,8 @@ class ImgDetectionExtended(dai.Buffer):
         """
         if not isinstance(value, float):
             raise TypeError("Confidence must be a float.")
-        if value < 0 or value > 1:
+        eps = 1e-3
+        if not (0 - eps <= value <= 1 + eps):
             raise ValueError("Confidence must be between 0 and 1.")
         self._confidence = value
 


jkbmrz:
What's the use case of conf. values being higher than 1?

HonzaCuhel:
The reason behind this is, so that when the model is quantized or the values are just badly rounded, the DAI script won't crash. Which is exactly what happened when I tested YOLOv7 tiny inference on a RVC4 device. The parser returned the following confidences scores: `[1.0000001192092896, 1.0000001192092896, 1.0000001192092896, 0.6650391221046448]`, which resulted in the script crash, as most of the values were `> 1`.

klemen1999 (commented):
Is there anywhere else we have a similar issue that we could handle inside this PR CC: @sokovninn  @aljazkonec1

kkeroo:
Can we rather clip the values to be between 0 and 1?

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -83,7 +83,8 @@ def confidence(self, value: float):
         """
         if not isinstance(value, float):
             raise TypeError("Confidence must be a float.")
-        if value < 0 or value > 1:
+        eps = 1e-3
+        if not (0 - eps <= value <= 1 + eps):
```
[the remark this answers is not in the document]
HonzaCuhel: Yeah, that sounds good. How big acceptance range do we want to set, do we want to also use the [-2, 2] interval or smth smaller, let's say [-1.5, 1.5]?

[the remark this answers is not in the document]
sokovninn: The maximum value I obtained from the quantized YOLO-World model is 1.0187655687332153, which does not pass this test. Should the eps value be increased?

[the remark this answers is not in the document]
klemen1999: > do we want to also use the [-2, 2] interval or smth smaller, let's say [-1.5, 1.5]?

The [-2,2] we have in the luxonis-ml is from different domain, here it should be much smaller. Based on YOLO-World perhaps [-0.1, 1.1] makes sense?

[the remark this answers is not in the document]
HonzaCuhel: Updated

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -83,8 +86,12 @@ def confidence(self, value: float):
         """
         if not isinstance(value, float):
             raise TypeError("Confidence must be a float.")
-        if value < 0 or value > 1:
+        if value < -0.1 or value > 1.1:
             raise ValueError("Confidence must be between 0 and 1.")
+        if not (0 <= value <= 1):
+            value = max(0, min(1, value))
+            logging.warning("Confidence value was clipped to [0, 1].")
```
sokovninn, line 93: Can we accept that this warning might be raised too frequently (case of YOLO-World)?

aljazkonec1:
> Is there anywhere else we have a similar issue that we could handle inside this PR CC: @sokovninn @aljazkonec1
Yes, Keypoints message also has similar clipping.

depthai_nodes/ml/messages/img_detections.py
HonzaCuhel: @klemen1999 do we want to keep the logging for each value, or skip it?

klemen1999: 2 points:
- in case of depthai-nodes we don't (yet) use logger so let's go with regular warnings
- And agree, we don't want to make this too frequent but IMO it's also valuable for user to know that this is not actually 1:1 output of the model and is being clipped. Does this really happen a lot?

klemen1999:
> Yes, Keypoints message also has similar clipping.

Let's also include same checks and clips there inside this PR as well then.

depthai_nodes/ml/messages/img_detections.py
HonzaCuhel: For obj det. yolos I'd say that often, but I don't know about the YOLO-World, @sokovninn how often does this happen for YOLO-World?

klemen1999: We can ty to get log level from Depthai and print this warning if log level is high enough ie. omit it in "normal use"

sokovninn: @HonzaCuhel when I tried it on a video example, it was about 1/3 of all frames.

klemen1999: Additions in [dd6461a](https://github.com/luxonis/depthai-nodes/pull/166/commits/dd6461aa278a99650b3315e5a8cee0a8a455c039): 
- Made this kind of logging at INFO level
- Level of logging controlled by env variable `DEPTHAI_NODES_LEVEL` (affects only dai-nodes logging), `DEPTHAI_LEVEL` (also used by DAI logger) or if both unset `WARN` level is used
This makes it so by default you shouldn't be spammed by this messages but you still have the option to see them if you want.

klemen1999 (approved):

PR #167, opened 2025-01-28
Tests: Docker support.

kkeroo:
This PR moves HIL execution to docker.

We have two images in our registry: one for e2e tests and one for integration tests. Image is pulled on the testbed itself. The tests run in the docker container completely. We are technically not limited anymore on specific testbeds but still we have some limitations because all testbeds do not include fully-pipeline-ready devices.

To add in the future:
- join the both images into a single one and optimize its size.


klemen1999 (commented):
LGTM, let's wait for tests to get picked up and pass. After we should clean up the commented code. 
Do we also want to update the e2e workflow?

kkeroo:
Yeah, I will clean up the code this is first attempt with workflow from the repo. I just want to make sure it works and passes. Then I will also push E2E changes - I have them locally.

klemen1999 (approved):
LGTM

PR #169, opened 2025-01-30
Feat: Depth Color Transform

dominik737:
`DepthColorTransform` node moved from experiments. Tests implemeted.


kkeroo (commented):
Won't "visualization stuff" be handled in the Visualizer?

dominik737:
> Won't "visualization stuff" be handled in the Visualizer?

No concrete plans. This would likely only be done if the bandwidth for depth (required for PCL) isn't possible to sufficiently compress. And there would be need to save the bandwidth so the _disparity -> PCL_ & _disparity -> colored disparity_ would be done on FE.

depthai_nodes/nodes/depth_color_transform.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        self._max_disparity = max_disparity
+
+    def process(self, disparity_frame: dai.Buffer) -> None:
+        """Processes incoming disparity frames and converts them to colored frames.
+
+        @param disparity_frame: Input disparity/depth frame
+        @type disparity_frame: depthai.ImgFrame
+        """
+        assert isinstance(disparity_frame, dai.ImgFrame)
```
klemen1999, line 62: Should this check be already in the function definition? So disparity_frame: `dai.ImgFrame` instead of `dai.Buffer`

klemen1999 (approved):
LGTM

depthai_nodes/nodes/depth_color_transform.py
dominik737: I believe the issue that we had was the `StereoDepth.disparity` does not have `dai.ImgFrame` as possible output. Which lead to inability to connect the two nodes straight. Can test again if it is still case in alpha12.

PR #170, opened 2025-01-30
Ref: Parsing Neural Network

dominik737:
Refactoring of `ParsingNeuralNetwork` featuring:

- Build method signatures fix to match DAI allowed signatures
- Option to use strongly typed `getParser` method
- Add `numShaves` parameter to `setNNArchive` method to match DAI allowed signature

### DIFF
--- a/depthai_nodes/parsing_neural_network.py
+++ b/depthai_nodes/parsing_neural_network.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import Dict, Union, overload
+from typing import Dict, Optional, Union, overload
 
 import depthai as dai
 
@@ -61,7 +61,7 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         self,
         input: dai.Node.Output,
         nn_source: Union[dai.NNModelDescription, dai.NNArchive, str],
-        fps: int = None,
+        fps: Optional[int] = None,
     ) -> "ParsingNeuralNetwork":
         """Builds the underlying NeuralNetwork node and creates parser nodes for each
         model head.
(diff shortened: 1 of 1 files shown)


depthai_nodes/parsing_neural_network.py
```diff
@@ -41,27 +43,36 @@ def __init__(self, *args, **kwargs) -> None:
...
     ) -> "ParsingNeuralNetwork":
         ...
 
     def build(
         self,
-        input: dai.Node.Output,
+        input: Union[dai.Node.Output, dai.node.Camera],
         nn_source: Union[dai.NNModelDescription, dai.NNArchive, str],
-        fps: int = None,
+        fps: Optional[float] = None,
```
jkbmrz, line 75: shouldn't `fps` be `int`?

jkbmrz (approved):
LGTM. Just a small comment on FPS dtype

kkeroo (approved):

depthai_nodes/parsing_neural_network.py
dominik737: I was looking at DAI and they want float parameter, so I preserved it such.

klemen1999: I confirmed with DepthAI, requesting float fps (e.g. 2.5) actually works so we can keep as suggested

depthai_nodes/parsing_neural_network.py
```diff
@@ -160,13 +178,18 @@ def setModelPath(self, modelPath: Path) -> None:
         """Sets the model path of the NeuralNetwork node."""
         self._nn.setModelPath(modelPath)
 
-    def setNNArchive(self, nnArchive: dai.NNArchive) -> None:
+    def setNNArchive(
+        self, nnArchive: dai.NNArchive, numShaves: Optional[int] = None
+    ) -> None:
```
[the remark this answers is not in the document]
dominik737: There is no return type on `setNNArchive`.

PR #172, opened 2025-02-05
Fix: Global Logging

dominik737:
## Purpose
Previous setup of the logging module led to duplicit logs when using the Python `logging` module along the `depthai-nodes` package as a dependency. 

## Specification
To get the package logger use the `get_logger` method from `depthai_nodes.utils`.

All the files and classes that were using the logger were changed to use the package logger.

## Testing & Validation
Was tested on simple script, where a custom logger was configured and log message from `depthai-nodes` was triggered. The logs were not duplicated by `depthai-nodes` logger. The logs from `depthai-nodes` were successfully logged.

### DIFF
--- a/depthai_nodes/ml/messages/img_detections.py
+++ b/depthai_nodes/ml/messages/img_detections.py
@@ -1,4 +1,3 @@
-import logging
 from typing import List, Tuple
 
 import depthai as dai
@@ -13,8 +12,7 @@ from depthai_nodes.ml.helpers.constants import (
 )
 from depthai_nodes.ml.messages.keypoints import Keypoint
 from depthai_nodes.ml.messages.segmentation import SegmentationMask
-
-logger = logging.getLogger(__name__)
+from depthai_nodes.utils import get_logger
 
 
 class ImgDetectionExtended(dai.Buffer):
@@ -43,6 +41,7 @@ class ImgDetectionExtended(dai.Buffer):
         self._label: int = -1
         self._label_name: str = ""
         self._keypoints: List[Keypoint] = []
+        self._logger = get_logger(__name__)
 
     @property
     def rotated_rect(self) -> dai.RotatedRect:
@@ -89,7 +88,7 @@ class ImgDetectionExtended(dai.Buffer):
             raise ValueError("Confidence must be between 0 and 1.")
         if not (0 <= value <= 1):
             value = max(0, min(1, value))
-            logger.info("Confidence value was clipped to [0, 1].")
+            self._logger.info("Confidence value was clipped to [0, 1].")
 
         self._confidence = value
 
--- a/depthai_nodes/ml/messages/keypoints.py
+++ b/depthai_nodes/ml/messages/keypoints.py
@@ -1,11 +1,9 @@
-import logging
 from typing import List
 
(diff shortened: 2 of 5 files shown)


depthai_nodes/ml/messages/img_detections.py
```diff
@@ -43,6 +41,7 @@ def __init__(self):
         self._label: int = -1
         self._label_name: str = ""
         self._keypoints: List[Keypoint] = []
+        self._logger = get_logger(__name__)
```
klemen1999, line 44: Does it make sense to have the logger be part of the class? To me it makes more sense to just get it outside of the class . Or would this conflict with getting the correct logger then?

klemen1999 (commented):
Generally LGTM

depthai_nodes/ml/messages/img_detections.py
dominik737: No conflict with getting the correct logger. In the yolo it is used outside of a class. When used inside a class it makes sense to me to have it rather as a class variable.

Classes should be a groupping all related variables. On the contrary class should not interact with variables outside of it's scoped. All dependent variables should be pass via constructor.

kkeroo (approved):

depthai_nodes/ml/messages/img_detections.py
```diff
@@ -89,7 +88,7 @@ def confidence(self, value: float):
             raise ValueError("Confidence must be between 0 and 1.")
         if not (0 <= value <= 1):
             value = max(0, min(1, value))
-            logger.info("Confidence value was clipped to [0, 1].")
+            self._logger.info("Confidence value was clipped to [0, 1].")
```
kkeroo, line 91: This could lead to spamming the log if many confidences are over 1. Is this visible in the stdout? Or only when we set DEPTHAI_NODES_LEVEL?

One way we can solve this is in the parser. We can have a flag indicating that we already printed this warning.

klemen1999: We've put it at the `info` level so most of the time you won't see this logs because the default level is `warn`.  Whereas if you would explicitly change the log level to something more precise then IMO you want to have print every time so it can help you debug.

kkeroo: Makes sense, thanks for clarification.

depthai_nodes/ml/messages/img_detections.py
klemen1999: Don't have a strong opinion but I think (at least in python) you setup the logger at module level not at class instance level (at least that is what I see in other usage examples). If nothing else getting it for each instance adds memory usage which is not needed.

dominik737: Depends on the package and how much you want to keep object oriented practices. For example SQL Alchemy uses logger on classes. I also prefer it that way. Truth is that with logger on classes you create the logger based on the class name, not the file name (`__name__`). Advantage is that each class has separate logger, which allows for more fine-grained control. The allocation cost is just negligible in a high-level language as Python. So it's up to you whether to create the logger based on the filename (`__name__`) or create it in a class based off the class name.

PR #173, opened 2025-02-09
Fix/restructure

klemen1999:
## Purpose

Restructures project structure so that the classes are accessable in similar pattern as with DepthAI.

## Specification

Moved all messages and message creators under `message`. Nodes are now under `node` (for now only parsers are in their own folder).

Previously:
```python
from depthai_nodes import ParsingNeuralNetwork
from depthai_nodes import Classifications
from depthai_nodes import ClassificationParser
```

Now:
```python
from depthai_nodes.node import ParsingNeuralNetwork
from depthai_nodes import Classifications
from depthai_nodes.node import ClassificationParser
```

## Deployment Plan

This is a breaking change so when merged in all scripts (e.g. depthai-experiments) relying on unreleased, main branch of dai-nodes will faill because of import errors. So before merge we should have PRs on other relevant projects that depend on this one already made.

## Testing & Validation

Tests were adjusted for the new structure.


dominik737 (approved):
Looks good. As was suggested by Petr in the future it would be nice to get rid of the `utils` and alike folders in the repo. Also using the constructor instead of the functional creators would be nice in the future.

For now it's good that both these things that we would like to change in the future are quite internal to the library. Thus no big breaking changes affecting users will be introduced.

jkbmrz (approved):
Generally LGTM. However, I'm wondering what’s the reasoning behind having the `node` directory at the top level of `depthai_nodes`? It feels a bit odd. Would it make sense to move `parsers` to the top level and rename `node` to `helpers` instead? That way, the imports would be more intuitive:

- depthai_nodes.message
- depthai_nodes.parsers
- depthai_nodes.helpers

Additionally, I’d recommend using plural forms for module names (e.g., message → messages, node → nodes) since the modules contain multiple related classes or functions.

klemen1999:
@jkbmrz 
Main reason is that we want to "mimic" how depthai already does it so we have same pattern of accesing "same things". And how DAI does it is: Messages are importable top level like `depthai.ImgDetections()` and all nodes are importable like `depthai.node.NeuralNetwork()`. 
And same reasoning goes for plural vs singular - DAI has singular so we want to be inline with them.

kkeroo (approved):
LGTM. One note: inside `node` folder we have `parsers` folder where parsers but other "host nodes" like DepthMerger are left in the `node` folder. In the future we should probably move them to separate subfolder when there will be a lot of such nodes?

dominik737:
@kkeroo
Yes, the idea is that now there are a few nodes and no meaningful groups can be formed yet. When we have more nodes we should be able to group them into subfolders.

PR #175, opened 2025-02-11
Fix/yoloextendedparser classes keypoints

klemen1999:
## Purpose

Previously NMS for keypoints was hardcoded to COCO person keypoints (17 points) which caused it to fail for custom models (e.g. [barcode](https://hub.luxonis.com/ai/models/75edea0f-79c9-4091-a48c-f81424b3ccab?view=page) model on HubAI). This PR fixes this.

## Specification

Changed NMS to remove some checks which caused other models to fail. Now we also apply sigmoid on kpt confidences by default (this can be changed down the line with new NNArchive/parser argument)

No breaking changes.

### DIFF
--- a/depthai_nodes/ml/parsers/utils/yolo.py
+++ b/depthai_nodes/ml/parsers/utils/yolo.py
@@ -337,7 +337,9 @@ def parse_yolo_output(
         )
     else:
         # Keypoints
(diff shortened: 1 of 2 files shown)


kkeroo (approved):

ptoupas (approved):
LGTM

PR #177, opened 2025-02-26
Set label names in detection parsers.

kkeroo:
## Purpose
We have `label_name` in the `ImgDetectionExtended` message but we do not set this field in the parsers. Setting this will allow the user to extract the class with `detection.label_name`.

## Specification
I adjusted `YOLOExtendedParser` and `DetectionParser` to get the class names from nn archive inside the `build` method. Relevant detection parsers are also adjusted.

No breaking changes.

## Deployment Plan
No plan.

## Testing & Validation
It was tested on all affected parsers. For yolo versions I tested pose estimation model and YOLOP.

### DIFF
--- a/depthai_nodes/ml/parsers/mediapipe_palm_detection.py
+++ b/depthai_nodes/ml/parsers/mediapipe_palm_detection.py
@@ -68,6 +68,7 @@ class MPPalmDetectionParser(DetectionParser):
         self.iou_threshold = iou_threshold
         self.max_det = max_det
         self.scale = scale
+        self.label_names = ["Palm"]
 
     def setOutputLayerNames(self, output_layer_names: List[str]) -> None:
         """Sets the output layer name(s) for the parser.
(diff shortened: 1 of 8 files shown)


klemen1999 (approved):

depthai_nodes/ml/parsers/mediapipe_palm_detection.py
```diff
@@ -194,9 +195,26 @@ def run(self):
...
+                label_names = [self.label_names[label] for label in labels]
+                detections_msg = create_detection_message(
+                    bboxes=bboxes,
+                    scores=scores,
+                    angles=angles,
+                    keypoints=points,
+                    labels=labels,
+                    label_names=label_names,
+                )
+            else:
```
jkbmrz, line 210: Do we really need this `else` statement here? Shouldn't we rather enforce the `self.label_names` to be a non-empty list or `None` (both are accepted by `create_detection_message`)

depthai_nodes/ml/parsers/scrfd.py
```diff
@@ -213,9 +214,23 @@ def run(self):
...
+            if self.label_names:
+                label_names = [self.label_names[label] for label in labels]
+                message = create_detection_message(
+                    bboxes=bboxes,
+                    scores=scores,
+                    labels=labels,
+                    label_names=label_names,
+                    keypoints=keypoints,
+                )
+            else:
```
jkbmrz, line 229: Similar comment as above, `else` statement not needed.

jkbmrz (approved):
LGTM, just a small stylistic suggestion to remove the `else` statements (I don't think it's needed if we properly enforce the parameter values)

depthai_nodes/ml/parsers/mediapipe_palm_detection.py
kkeroo: Mhm good catch! No need fo if-else. Removed in [791dac2](https://github.com/luxonis/depthai-nodes/pull/177/commits/791dac2cfb3e7e8fa03731f39e12a42abb4f7f1d).

depthai_nodes/ml/parsers/scrfd.py
kkeroo: Removed in [791dac2](https://github.com/luxonis/depthai-nodes/pull/177/commits/791dac2cfb3e7e8fa03731f39e12a42abb4f7f1d).

PR #184, opened 2025-03-06
E2E tests extension.

kkeroo:
## Purpose
We want to have test support for different releases with the preselected models that are known to work with the dai-nodes release.

## Specification
We extend Dockerfile and update the docker image on GHCR with the depthai-nodes-version parameter. We can pass the release tag (like `v0.1.2-alpha`) or branch name like before. Therefore we also introduce the same argument for the `main.py` file. This argument is then used to check if we have a list stored in a json that belogs to the given key (dai-nodes-version). The idea if json file is to have a list of preselected models that for for every release. For the upcomming release (v.0.2.0-alpha) I added all the available models because they all work. I also added the "version" `tests/e2e-extension` which is the name of this branch and is used only for testing. I will remove this after the release.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested locally and using GH actions. [Here](https://github.com/luxonis/depthai-nodes/actions/runs/13702189719/job/38318580056) is a test with the provided branch (note that I triggered the test with flag -p YuNetParser to finish the test faster instead of -all) and [here](https://github.com/luxonis/depthai-nodes/actions/runs/13706281616/job/38333823185) is the test with provided release tag that runs on three predefined models from json file.


tests/end_to_end/Dockerfile
```diff
@@ -26,4 +26,15 @@ RUN pip install -r requirements-dev.txt
 WORKDIR /app/depthai-nodes/tests/end_to_end
 
 # Run the setup script and the tests
-ENTRYPOINT ["bash", "-c", "git checkout main && git pull && git checkout $BRANCH && pip install --extra-index-url $LUXONIS_EXTRA_INDEX_URL depthai==$DEPTHAI_VERSION && source <(python setup_camera_ips.py) && python main.py $FLAGS"]
\ No newline at end of file
+# ENTRYPOINT ["bash", "-c", "git checkout main && git pull && git checkout $BRANCH && pip install --extra-index-url $LUXONIS_EXTRA_INDEX_URL depthai==$DEPTHAI_VERSION && source <(python setup_camera_ips.py) && python main.py $FLAGS"]
```
klemen1999, line 29: We can remove the comment

tests/end_to_end/main.py
```diff
@@ -34,13 +35,32 @@ def main():
...
+    )
 
     args = arg_parser.parse_args()
     nn_archive_path = args.nn_archive_path  # it is a list of paths
     model = args.model
     run_all = args.all
     parser = args.parser
     rvc_platform = "both" if args.platform == "" else args.platform
+
+    print()
```
klemen1999, line 52: redundant print?

tests/end_to_end/supported_models.json
```diff
@@ -0,0 +1,127 @@
...
+        "yolov6-large:r2-coco-640x352",
+        "yolov6-nano:r2-adjusted-coco-512x288",
+        "yolov6-nano:r2-coco-512x288",
+        "superanimal-landmarker:256x256",
+        "mediapipe-palm-detection:192x192",
+        "mediapipe-face-landmarker:192x192",
+        "scrfd-face-detection:10g-640x640"
+    ],
+    "v0.1.2-alpha": [
+        "superanimal-landmarker:256x256",
```
klemen1999, line 118: Let's populate with real values, so all from v0.2.0-alpha - barcode model

tests/end_to_end/supported_models.json
```diff
@@ -0,0 +1,127 @@
...
+        "mediapipe-palm-detection:192x192",
+        "mediapipe-face-landmarker:192x192"
+    ],
+    "tests/e2e-extension": [
```
klemen1999, line 122: Can be removed right before merge of this PR

klemen1999 (approved):
LGTM thanks, left just some nitpicks

klemen1999:
CC: @danilo-pejovic to double check if this integration works for BOM test. If possible I would adjust the [run_e2e_tests.sh](https://github.com/luxonis/depthai-nodes/blob/main/tests/end_to_end/run_e2e_tests.sh#L30) already in this PR to replace setting -m flag to passing in the depthai-nodes tag/branch

tests/end_to_end/Dockerfile
kkeroo: Yea I guess we can. I left it there intentionally so we could easily revert. Removed in [a3ffcc1](https://github.com/luxonis/depthai-nodes/pull/184/commits/a3ffcc1a02c3ce11081b00db18c034eef3a8805e).

tests/end_to_end/main.py
kkeroo: Yes and no, I wanted to print a new line so the print statements in the console are more visible. I removed the line and added the \n char.

tests/end_to_end/supported_models.json
kkeroo: I acctually removed the v0.1.2 because it wont work since the testing script in that version is not compatible with this. So the first version that will work is 0.2.0.

tests/end_to_end/supported_models.json
kkeroo: Yeah, lets leave it so @danilo-pejovic can test the BOM by giving the "tests/e2e-extension" as the "release".

PR #185, opened 2025-03-08
Import everything from node

HonzaCuhel:
## Purpose
Import everything from .node in the `depthai_nodes/__init__.py` to simplify the imports when using the package.

## Specification

None

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested locally

### DIFF
--- a/depthai_nodes/__init__.py
+++ b/depthai_nodes/__init__.py
@@ -1,6 +1,7 @@
 from .constants import *
 from .logging import setup_logging
 from .message import *
+from .node import *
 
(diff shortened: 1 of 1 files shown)


klemen1999:
We purposefully didn't implement it like that is because we want to copy how DAI does it so the API is more uniform across our stack. So DAI for example has nodes accessed like `dai.node.Camera` where as messages and types like `dai.ImgDetections`. So this change in dai-nodes is intentional and breaking (thus release 0.2.0).

PR #186, opened 2025-03-10
Integration tests: Remove HIL dependency

kkeroo:
## Purpose
Since we are only testing parsers we can avoid running the whole DAI pipeline with device and create mock pipeline and DAI components instead. This way we will free up the testbed usage. Additionally, we also want to run the test for the parser for long time to see if there are no leaks etc. 

## Specification
We create Mock Pipeline and ThreadedHostNode with custom queues, inputs, and outputs. The dai classes are replaced with mock classes so we can create `ParserGenerator` with `pipeline.create`. Parameter `duration` is added to control how long the input queue will send the same NNData to the parser. I couldn't find a better option than having this duration inside the queue itself.

None / not applicable

## Deployment Plan
You can trigger the tests manually and set the duration parameter. However, tests are still fired inside each PR with duration set to `2` seconds.

## Testing & Validation
Tested locally and in CI. [Here](https://github.com/luxonis/depthai-nodes/actions/runs/13761141284) is the run.


tests/conftest.py
```diff
@@ -13,3 +13,9 @@ def pytest_addoption(parser):
         "--models", action="store", default="", help="Model slug from the ZOO."
     )
     parser.addoption("--parsers", action="store", default="", help="Parsers to test.")
+    parser.addoption(
+        "--duration",
+        action="store",
+        default="",
```
klemen1999, line 19: Any specific reason we have action=store specifically set in these arguments? Afaik this is a default action anyway.
And should we rather explicitly set types and default values? In this case perhaps type=int, default=10 makes sense?

klemen1999 (approved):
Generally LGTM

tests/conftest.py
kkeroo: Yes, we can. Added [88c3cff](https://github.com/luxonis/depthai-nodes/pull/186/commits/88c3cff2bbc0d3214466da2cfd0d2284cae16c69). However, this is only used internally for pytest. You provide the parameters to the `main.py` file [here](https://github.com/luxonis/depthai-nodes/blob/88c3cff2bbc0d3214466da2cfd0d2284cae16c69/tests/integration_tests/main.py#L28).

kkeroo:
I added the unittests for `ParserGenerator` and `ParsingNeuralNetwork`. In PNN we test `getParser`, `getOutput`, `setNNArchive` functions along with the `build` function. Other setters/getters are dai functions.

PR #187, opened 2025-03-11
Updated DetectionParser.

kkeroo:
## Purpose
By adding Mobile object localizer model to our ZOO we need a parser for it. I think we can use `DetectionParser` and just define the `run` method for `general simple` object detector - get bboxes and scores and compute NMS and return the message.

## Specification
We add the `run` method to the `DetectionParser` class so it can be used as standalone parser.

Does not affect other detection parsers as they all override `run` method.

## Deployment Plan
None / not applicable

## Testing & Validation
Tested with general example in dai-experiments on Mobile obj. localizer, YuNet, SCRFD, PalmDetection models.

### DIFF
--- a/depthai_nodes/node/parsers/detection.py
+++ b/depthai_nodes/node/parsers/detection.py
@@ -139,7 +139,7 @@ class DetectionParser(BaseParser):
 
             for layer in layers:
                 tensor: np.ndarray = output.getTensor(layer, dequantize=True)
-                if 4 in tensor.shape:
+                if tensor.shape[-1] == 4 and len(tensor.shape) != 1:
                     bboxes = tensor
                 else:
                     scores = tensor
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
LGTM, I think this makes sense to extend and hopefully it should be applicable to some other simple detection models in the future

depthai_nodes/node/parsers/detection.py
```diff
@@ -109,3 +120,51 @@ def build(self, head_config) -> "DetectionParser":
...
+                raise ValueError(
+                    f"Expected 2 output layers, got {len(layers)} layers. Please use different parser or create a new one."
+                )
+
+            bboxes = None
+            scores = None
+
+            for layer in layers:
+                tensor: np.ndarray = output.getTensor(layer, dequantize=True)
+                if 4 in tensor.shape:
```
jkbmrz, line 142: In case the model produces only 4 detections the scores tensor would have shape `(4,)`. I assume it would be incorrectly set as `bboxes` variable in that case?

jkbmrz (commented):
LGTM, just adding a small comment on validation.

jkbmrz (approved):

PR #188, opened 2025-03-14
fix: image creator function

jkbmrz:
## Purpose
Outputs of the `ImageOutputParser` are not visualized correctly in DAI Visualizer due to a bug in the `create_image_message()` creator.

## Specification
Changes in `create_image_message()`:
- Use `.setCvFrame()` instead of `.setData()` when creating the `dai.ImgFrame`,
- add an argument (`img_frame_type`) to adjust the output `ImgFrame.Type` to the running platform.

Changes in `ImageOutputParser`:
- extract running platform and pass the correct `ImgFrame.Type` when calling the `create_image_message()`.

None

## Deployment Plan
None

## Testing & Validation
The changes were tested by running the `generic-example` both on **RVC2** and **RVC4** for the following models:
- `rt-super-resolution`
- `esrgan` (RVC4 only)
- `zero-dce`


jkbmrz:
To discuss: `ImageOutputParser` currently always ouputs `dai.ImgFrame` of type `BGR888i`. This works for the simple visualizations (e.g. `generic-example`) but might we want to adjust it to the device the pipeline is running on (e.g. if input to the model is `BGR888p`, that the parser also outputs `BGR888p` instead of `BGR888i` as is the case now).

kkeroo (approved):

kkeroo:
> but might we want to adjust it to the device the pipeline is running on

We can since it is probably easy addition (`self.getParentPipeline()...`)

PR #189, opened 2025-03-17
Tests: Stability tests for host nodes.

kkeroo:
## Purpose
Besides stability tests for parsers we also need stability tests for host nodes and hence the PR is created. Tests for `Tiling` and `TilesPatcher` nodes are also added.

## Specification
We adjust unittests for host nodes so they also accept optional fixture `duration` which can be passed to set the duration of the stability tests to run.

None / not applicable

## Deployment Plan
It is already included in the CI.

## Testing & Validation
Tested locally.


tests/unittests/test_nodes/test_depth_color_transform.py
```diff
@@ -30,16 +37,27 @@ def empty_frame(disp_shape):
...
+def test_dimensions_type(disp_frame, disp_shape, duration):
     color_transform = DepthColorTransform()
     q = color_transform.out.createOutputQueue()
     color_transform.process(disp_frame)
     output = q.get()
     assert isinstance(output, dai.ImgFrame)
     assert output.getCvFrame().shape == (*disp_shape, 3)
 
+    if not duration:
+        return
```
jkbmrz, line 49: Just to double check, so if `duration` is `0` or `None`, the function will stop running here? Nitpicking, but to me this would read better if written as:
```
if duration:
... (remainder of the code below)
```

tests/unittests/test_nodes/test_depth_color_transform.py
```diff
@@ -30,16 +37,27 @@ def empty_frame(disp_shape):
...
+
+    start_time = time.time()
+
+    while time.time() - start_time < duration:
```
jkbmrz, line 53: Before going into loop, we should probably check that the `duration` is not negative. Otherwise, this will loop infinitely.

jkbmrz (approved):
LGTM, just two small suggestions.

kkeroo:
The integration tests are renamed to stability tests now. README is updated. Test for `DetectionParser` is added. [48c8453](https://github.com/luxonis/depthai-nodes/pull/189/commits/48c845342e74b450374d8ffe06a294100e59c714).

klemen1999 (approved):

PR #192, opened 2025-03-18
feat: Porting host nodes from depthai-experiments

jkbmrz:
## Purpose
Porting over some host nodes from [depthai-experiments](https://github.com/luxonis/depthai-experiments/tree/gen3) with potential for common usage.

## Specification
The following host nodes were ported:
- `ImgFrameOverlay`, 
- `ImgDetectionsBridge`,
- `ImgDetectionsFilter`.

The following host nodes were modified:
-  `DepthColorTransform` (made more general, renamed to `ApplyColormap`)

Changes in `DepthColorTransform` node have effect on some depthai-experiments. Addressed in [#631](https://github.com/luxonis/depthai-experiments/pull/631)

## Deployment Plan
None

## Testing & Validation
All 4 nodes were tested by running custom demos on both **RVC2** and **RVC4**:

- `yolov6-nano:r2-coco-512x288` (`dai.ImgDetections`)
- `yolov8-instance-segmentation-nano:coco-512x288` (`ImgDetectionsExtended`)
- `midas-v2-1:small-256x192` (`Map2D`)


klemen1999 (commented):
Let's also add unittests for both of the nodes

klemen1999 (commented):
Generally comment: Let's add unittests for every new added node. You can check ones already there for reference and/or ping @kkeroo for help on how to make mock pipelines

jkbmrz:
cc @kkeroo to please check the unittests more thoroughly.

depthai_nodes/node/img_detections_filter.py
```diff
@@ -0,0 +1,133 @@
...
+
+            if self._confidence_threshold is not None:
+                if detection.confidence < self._confidence_threshold:
+                    continue
+
+            filtered_detections.append(detection)
+
+        msg_new.detections = filtered_detections[
+            : self._max_detections
+        ]  # TODO: sort detections by confidence before subsetting?
```
klemen1999, line 131: Maybe for this we add another standalone parameter `sort: bool` and the sorting would be another form of filtering that would (if set to True) explicitly happen before max_detection filtering? And would be good to note down in the docstring what is the order of this operations

jkbmrz:
PS. Next to porting the nodes, I've also started implemented here a message copying mechanism (see the `depthai_nodes/node/utils/copy_message.py`). Use case is that some host nodes need to modify a message but don't want to change it at source because it might also be used elsewhere. `depthai` doesn't support message copying unfortunately, so these messages need to be copied manually. For `depthai-nodes` messages, I've started adding to them the `.copy()` methods to support the message copying. I've only implemented this for the messages utilized in the nodes ported in this PR but we can extend this copying mechanism in a further PR. Thoughts?

depthai_nodes/node/img_frame_overlay.py
```diff
@@ -0,0 +1,100 @@
...
+        self.SetBackgroundWeight(background_weight)
+
+        try:
+            self._platform = (
+                self.getParentPipeline().getDefaultDevice().getPlatformAsString()
+            )
+        except AttributeError:
+            self._platform = None  # so that it doesn't crash when running unittests
+
+    def SetBackgroundWeight(self, background_weight: float) -> None:
```
klemen1999, line 32:
```suggestion
    def setBackgroundWeight(self, background_weight: float) -> None:
```

depthai_nodes/node/img_frame_overlay.py
```diff
@@ -0,0 +1,100 @@
+import cv2
+import depthai as dai
+
+
+class ImgFrameOverlay(dai.node.HostNode):
```
klemen1999, line 5: Should we make this more geenric and have `frame1` and `frame2` naming instead of background and foreground? And then the weight would be called `alpha` which would specify weight of the first frame. Taking inspiration from OpenCV docs [here](https://docs.opencv.org/3.4/d2/de8/group__core__array.html#gafafb2513349db3bcff51f54ee5592a19)

depthai_nodes/node/utils/copy_message.py
```diff
@@ -0,0 +1,58 @@
...
+        img_dets_copy.setTransformation(img_dets.getTransformation())
+        return img_dets_copy
+
+    if isinstance(msg, dai.ImgDetection):
+        return _copy_img_detection(msg)
+    elif isinstance(msg, dai.ImgDetections):
+        return _copy_img_detections(msg)
+    else:
+        # TODO: define logic for copying other message types
+        raise TypeError(f"Copying of message type {type(msg)} is not supported.")
```
klemen1999, line 58: Does this currently fail on every custom dai-nodes message except ImgDetection and SegmentationMask?

depthai_nodes/node/utils/copy_message.py
klemen1999, line 1: Let's also add unittests for this `copy_message` function

klemen1999 (commented):
Generally LGTM, left some comments

depthai_nodes/node/utils/copy_message.py
jkbmrz: Yeah, it should I think. Other custom nodes currently miss the .copy() method but I'm proposing we add it to all of them.

klemen1999: I'm not opposed to adding it on all messages yeah, I think it could be quite useful to have in specific data manipulation scenarious

depthai_nodes/node/apply_colormap.py
```diff
@@ -0,0 +1,148 @@
...
+        self.setColormap(colormap_value)
+        self.setMaxValue(max_value)
+        self.setInstanceToSemanticMask(instance_to_semantic_mask)
+
+        try:
+            self._platform = (
+                self.getParentPipeline().getDefaultDevice().getPlatformAsString()
+            )
+        except AttributeError:
+            self._platform = None  # so that it doesn't crash when running unittests
```
kkeroo, line 45: We should handle this in mock classes and not set the pipeline to None. I think `getParentPipeline().getDefaultDevice().getPlatformAsString()` should work already. If not we can add the methods easily.

depthai_nodes/node/img_detections_filter.py
```diff
@@ -0,0 +1,133 @@
...
+
+        return self
+
+    def process(self, msg: dai.Buffer) -> None:
+        assert isinstance(
+            msg,
+            (
+                dai.ImgDetections,
+                ImgDetectionsExtended,
+            ),  # TODO: also allow dai.SpatialImgDetections?
```
kkeroo, line 106: Maybe we can already allow dai.SpatialImgDetections?

depthai_nodes/node/img_frame_overlay.py
kkeroo: Just make sure to specify which frame is on top and which bellow. So user doesnt have to check the code.

depthai_nodes/node/utils/copy_message.py
```diff
@@ -0,0 +1,58 @@
...
+    def _copy_img_detections(img_dets: dai.ImgDetections) -> dai.ImgDetections:
+        assert isinstance(img_dets, dai.ImgDetections)
+        img_dets_copy = dai.ImgDetections()
+        img_dets_copy.detections = [
+            _copy_img_detection(img_det) for img_det in img_dets.detections
+        ]
+        img_dets_copy.setSequenceNum(img_dets.getSequenceNum())
+        img_dets_copy.setTimestamp(img_dets.getTimestamp())
+        img_dets_copy.setTimestampDevice(img_dets.getTimestampDevice())
+        img_dets_copy.setTransformation(img_dets.getTransformation())
```
kkeroo, line 49: Is this all message metadata? Hopefully we dont miss any

kkeroo (changes requested):
Left some comments.

depthai_nodes/node/img_detections_filter.py
jkbmrz: Do you maybe know which model produces those so that I could test it out?

depthai_nodes/node/utils/copy_message.py
jkbmrz: Yeah this is a bit slippery. We can ask DAI team to check once we have everything ported in a separate PR.

depthai_nodes/node/img_detections_filter.py
kkeroo: No one, you can test it with `SpatialYOLONetwork`. IIRC it has the same attrs. as imgDetections with additional attrs. for spatial coordinates.

depthai_nodes/node/img_frame_overlay.py
```diff
@@ -0,0 +1,100 @@
...
+
+        overlay = dai.ImgFrame()
+        overlay.setCvFrame(
+            overlay_frame,
+            (
+                dai.ImgFrame.Type.BGR888i
+                if self._platform == "RVC4"
+                else dai.ImgFrame.Type.BGR888p
+            ),
+        )
```
dominik737, line 96: This logic is repeated very frequently also with the
```
self._platform = (
    self.getParentPipeline().getDefaultDevice().getPlatformAsString()
)
```
We should make an abstract base class that would have this functionality. We are basically doing this because of the different NN input formats for the platforms. Possible name could be `NNImageSender`, but could be improved.

depthai_nodes/node/utils/copy_message.py
```diff
@@ -0,0 +1,58 @@
...
+        return copy.deepcopy(msg)  # First attempt: use deepcopy
+    except Exception:
+        pass  # If deepcopy fails, move to the next approach
+
+    try:
+        return msg.copy()  # Second attempt: use .copy() method
+    except AttributeError:
+        pass  # If .copy() method doesn't exist, move to the next approach
+
+    return _copy(msg)  # Last attempt: use a custom copy implementation
```
dominik737, line 25: Why this exact sequence? I would think that we would go from the most common and optimized to the least. The `deepcopy` is the most general way to copy objects. The `.copy()` and `_copy` are very specific thus IMO should be give priority.

I would propose:

1. `hasattr` check, if sucess then do `getattr`
2. try `_copy` there could also be flow such `_can_copy` -> `_copy`
3. last resort `deepcopy`

This should avoid the situation where specific objects with either `copy` method implemented or `_copy` possible are being copied with generic (and slow) `deepcopy`.

depthai_nodes/node/utils/copy_message.py
```diff
@@ -0,0 +1,58 @@
...
+
+    try:
+        return copy.deepcopy(msg)  # First attempt: use deepcopy
+    except Exception:
+        pass  # If deepcopy fails, move to the next approach
+
+    try:
+        return msg.copy()  # Second attempt: use .copy() method
+    except AttributeError:
+        pass  # If .copy() method doesn't exist, move to the next approach
```
dominik737, line 23: You can check with `hasattr` I think it's faster than catching exception, which is usually pretty slow.

dominik737 (approved):
Looks good, just some code suggestions.

depthai_nodes/node/img_detections_filter.py
jkbmrz: Added in [5e34740](https://github.com/luxonis/depthai-nodes/pull/192/commits/5e34740993ccacdfc98245452b64e6e7b098363f).

depthai_nodes/node/apply_colormap.py
jkbmrz: Added in [19b823f](https://github.com/luxonis/depthai-nodes/pull/192/commits/19b823f4ac595ee47964a663f57d6a515c9f89ca). @kkeroo could you please double check?

depthai_nodes/node/utils/copy_message.py
jkbmrz: Added in [c7eda44](https://github.com/luxonis/depthai-nodes/pull/192/commits/c7eda44faefed3d90f6ea0cd4db284e07339f2a0).

depthai_nodes/node/img_frame_overlay.py
jkbmrz: Added in [be6db22](https://github.com/luxonis/depthai-nodes/pull/192/commits/be6db22fbaf698d20b9b2b46958d078f615a1b12).

depthai_nodes/node/img_detections_filter.py
jkbmrz: Added in [be6db22](https://github.com/luxonis/depthai-nodes/pull/192/commits/be6db22fbaf698d20b9b2b46958d078f615a1b12).

depthai_nodes/node/utils/copy_message.py
jkbmrz: Added in [b2ea643](https://github.com/luxonis/depthai-nodes/pull/192/commits/b2ea64371be5090ca214a763d0cd42c831372f29).

depthai_nodes/node/img_frame_overlay.py
jkbmrz: Fixed in [4282518](https://github.com/luxonis/depthai-nodes/pull/192/commits/42825189d4e4043fd27f097c311274e3dfb5a149).

depthai_nodes/node/img_frame_overlay.py
jkbmrz: I agree. But I'd add it in a separate PR because this one is getting a bit too big.

jkbmrz:
Based on the comments above, these are the TODOs for the next PRs:

- extend message copying mechanism for all `depthai`(`-nodes`) messages (described [HERE](https://github.com/luxonis/depthai-nodes/pull/192#issuecomment-2761545477))
    - once implemented, ask the DAI team to review our copying mechanism for their messages (e.g. if we catch all attributes and metadata)
- make an abstract base class (e.g. `NNImageSender`) that handles NN input/output formats for different platforms (described [HERE](https://github.com/luxonis/depthai-nodes/pull/192#discussion_r2021554304))

tests/unittests/test_nodes/test_apply_colormap.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+
+
+@pytest.fixture(scope="session")
+def duration(request):
+    return request.config.getoption("--duration")
+
+
+@pytest.fixture
+def img_detections_extended():
```
kkeroo, line 40: This looks like its not used.

tests/unittests/test_nodes/test_copy_message.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+    else:
+        raise ValueError(f"Unsupported attribute type: {type(obj1)}")
+
+
+@pytest.mark.parametrize(
+    "message_creator",
+    inspect.getmembers(create_message, inspect.isfunction),
+)
+def test_message_copying(message_creator: Tuple[str, Callable], duration: int = 1e-6):
```
kkeroo, line 51: Copy message is not a host node and is not useful for stability tests since it is only a function. You can remove the duration parameter and while loop. Also move the test file outside the test_nodes folder, maybe test_messages?

kkeroo (commented):
LGTM, two small comments.

tests/unittests/test_nodes/test_copy_message.py
jkbmrz: Added in [a83e8a7](https://github.com/luxonis/depthai-nodes/pull/192/commits/a83e8a7405fbff151bc98c39a6e1c1989ba70b18) and [3814c20](https://github.com/luxonis/depthai-nodes/pull/192/commits/3814c20fe83c70f348cbd45f738e8c619ced7739).

depthai_nodes/node/apply_colormap.py
```diff
@@ -0,0 +1,145 @@
...
+        If True, converts instance segmentation masks to semantic segmentation masks.
+    arr : dai.ImgFrame or Map2D or ImgDetectionsExtended
+        The input message with a 2D array.
+    output : dai.ImgFrame
+        The output message for a colorized frame.
+    """
+
+    def __init__(
+        self,
+        colormap_value: int = cv2.COLORMAP_HOT,
```
klemen1999, line 28: Any speciific reason we have HOT and not JET as default? Usually JET is more used because the min and max values are less similar and is thus easier to spot the differece

depthai_nodes/node/apply_colormap.py
```diff
@@ -0,0 +1,145 @@
...
+    """A host node that applies a colormap to a given 2D array (e.g. depth maps,
+    segmentation masks, heatmaps, etc.).
+
+    Attributes
+    ----------
+    colormap_value : Optional[int]
+        OpenCV colormap enum value. Determines the applied color mapping.
+    max_value : Optional[int]
+        Maximum value to consider for normalization. If set lower than the map's actual maximum, the map's maximum will be used instead.
+    instance_to_semantic_mask : Optional[bool]
```
klemen1999, line 18: This is only relevant if input message type is ImgDetectionsExtended right? If so let's note this in the docstrings

depthai_nodes/node/img_detections_bridge.py
```diff
@@ -0,0 +1,120 @@
...
+        elif isinstance(msg, ImgDetectionsExtended):
+            msg_transformed = self._img_det_ext_to_img_det(msg)
+        else:
+            raise TypeError(
+                f"Expected dai.ImgDetections or ImgDetectionsExtended, got {type(msg)}"
+            )
+
+        msg_transformed.setTimestamp(msg.getTimestamp())
+        msg_transformed.setSequenceNum(msg.getSequenceNum())
+        # msg_transformed.setTransformation(msg.getTransformation()) # TODO: uncomment once .setTransformation() and .getTransformation() are supported
```
klemen1999, line 49: Is this a known DAI limitation that is in the works? Or limitation on dai-nodess message type side?

depthai_nodes/node/img_detections_bridge.py
```diff
@@ -0,0 +1,120 @@
+import depthai as dai
+
+from depthai_nodes import ImgDetectionExtended, ImgDetectionsExtended
+
+
+class ImgDetectionsBridge(dai.node.HostNode):
+    """Transforms the dai.ImgDetections to ImgDetectionsExtended object or vice versa".
```
klemen1999, line 7: Let's note that this conversion from ImgDetectionsExtended to ImgDetection loses information about segmentation, keypoints and rotation

depthai_nodes/node/img_detections_bridge.py
```diff
@@ -0,0 +1,120 @@
...
+        assert isinstance(img_det_ext, ImgDetectionsExtended)
+
+        img_dets = dai.ImgDetections()
+
+        detections_transformed = []
+        for detection in img_det_ext.detections:
+            detection_transformed = dai.ImgDetection()
+            detection_transformed.label = detection.label
+            detection_transformed.confidence = detection.confidence
+            if detection.rotated_rect.angle != 0:
```
klemen1999, line 102: I guess we are able to transform as an outer rect but we lose some information. I would maybe remove this error and rather use logger (check [here](https://github.com/luxonis/depthai-nodes/blob/db2ed1657c457d7fcdcb044f3bb1d1530f94830f/depthai_nodes/message/keypoints.py#L31) how to set it up) and logger.debug() a message every time on the start of this function what information is lost (eg. "Transformation from ImgDetectionsExtended to ImgDetections type will result in lose of keypoint, segmentation and rotated rectangle information if present in the original message" - maybe reworded nicer :) )

depthai_nodes/node/img_detections_filter.py
```diff
@@ -0,0 +1,158 @@
...
+
+import depthai as dai
+
+from depthai_nodes import ImgDetectionsExtended
+from depthai_nodes.node.utils import copy_message
+
+
+class ImgDetectionsFilter(dai.node.HostNode):
+    """Filters out detections based on the specified criteria. The order of operations:
+    filter by label/confidence -> sort -> subset.
```
klemen1999, line 11: Let's also mentioned that this is not done in place, that the returned mesage will be a copy of the original

klemen1999 (approved):
Generally LGTM, just left some smaller suggestions

depthai_nodes/node/img_detections_bridge.py
jkbmrz: Seems like this was added since the PR was opened and re-branching fixed this. Will remove the comment!

depthai_nodes/node/img_detections_filter.py
jkbmrz: Added in [b175847](https://github.com/luxonis/depthai-nodes/pull/192/commits/b175847897dc9b64ca6dacc95f24084af9886874).

depthai_nodes/node/img_detections_bridge.py
jkbmrz: Added in [3ce194a](https://github.com/luxonis/depthai-nodes/pull/192/commits/3ce194a93c800c8da69296967b416157912d6775).

depthai_nodes/node/img_detections_bridge.py
jkbmrz: Added in [7824801](https://github.com/luxonis/depthai-nodes/pull/192/commits/78248016d8534be3851a3b1d8798d1c097e7bd96).

depthai_nodes/node/img_detections_bridge.py
jkbmrz: Added in [670f3d7](https://github.com/luxonis/depthai-nodes/pull/192/commits/670f3d7a47e4408546d3e13036dc63e5c5b073ac).

depthai_nodes/node/apply_colormap.py
jkbmrz: Added in [6cdb50e](https://github.com/luxonis/depthai-nodes/pull/192/commits/6cdb50e5cfc7c1a36b3a76bf76974caf09a2744e).

depthai_nodes/node/apply_colormap.py
jkbmrz: [2d2457d](https://github.com/luxonis/depthai-nodes/pull/192/commits/2d2457d04f8b431a2c6c406ba61a6cd45221ea12)

PR #193, opened 2025-03-24
DetectionParser bug fix.

kkeroo:
## Purpose
We had a bug in DetectionParser because it was sending out the message only when there was a detection. Small fix also on E2E workflow.

## Specification
We send out empty message if there is no detections. In E2E worklow we explicitly set the requested platform to `rvc2 and rvc4`.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
None / not applicable


jkbmrz (approved):

.github/workflows/e2e_tests.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
...
+            hil --capabilities "depthai-core-hil" --platforms "rvc2 and rvc4" --wait --reservation-name $RESERVATION_NAME --before-docker-pull "echo ${{secrets.GHCR_PAT}} | docker login ghcr.io -u ${{secrets.GHCR_USER}} --password-stdin" --docker-image ghcr.io/luxonis/depthai-nodes-testing --docker-run-args "--env LUXONIS_EXTRA_INDEX_URL=${{secrets.LUXONIS_EXTRA_INDEX_URL}} --env DEPTHAI_VERSION=${{ github.event.inputs.depthai-version }} --env HUBAI_TEAM_SLUG=${{ secrets.HUBAI_TEAM_SLUG }} --env HUBAI_API_KEY=${{ secrets.HUBAI_API_KEY }} --env BRANCH=${{ github.event.inputs.depthai-nodes-version }} --env FLAGS=\"${{ github.event.inputs.additional-parameter }} -v ${{ github.event.inputs.depthai-nodes-version }}\""
```
klemen1999, line 52: HIL team mentioned that we should use `exec hil ...` (so appending `exec` before `hil` command) everywhere. Let's include this in this PR across all workflows

kkeroo: Ok, since we only use HIL in E2E now only this workflow is updated.

PR #194, opened 2025-03-24
Add TwoStageSync node

HonzaCuhel:
## Purpose
Adding `TwoStageSync` node

## Specification
- Adding `TwoStageSync` and unittests

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


kkeroo:
We support also Python 3.8 so some features are not available there like `|` in the typehints

depthai_nodes/message/detected_recognitions.py
```diff
@@ -0,0 +1,15 @@
+import depthai as dai
+from .img_detections import ImgDetectionsExtended
+
+
+class DetectedRecognitions(dai.Buffer):
+    def __init__(
+        self, detections: dai.ImgDetections | ImgDetectionsExtended, nn_data: list[dai.NNData]
```
kkeroo, line 7: Not supported in Python 3.8. Check also elsewhere.

kkeroo (changes requested):

depthai_nodes/message/detected_recognitions.py
```diff
@@ -0,0 +1,22 @@
+from typing import List, Union
+
+import depthai as dai
+
+from .img_detections import ImgDetectionsExtended
+
+
+class DetectedRecognitions(dai.Buffer):
```
jkbmrz, line 8: We probably want this to follow the format of other messages (adding docstrings, adding getter/setter methods, etc.).

depthai_nodes/node/detections_recognitions_sync.py
```diff
@@ -0,0 +1,161 @@
+import time
+from queue import PriorityQueue
+from typing import Dict, List, Optional, Union
+
+import depthai as dai
+
+from depthai_nodes import DetectedRecognitions, ImgDetectionsExtended
+
+
+class DetectionsRecognitionsSync(dai.node.ThreadedHostNode):
```
jkbmrz, line 10: Any reason why we want to have this inherit from `dai.node.ThreadedHostNode` instead of `dai.node.HostNode` (as is the case for other nodes)?

depthai_nodes/node/detections_recognitions_sync.py
```diff
@@ -0,0 +1,161 @@
...
+            float,
+            Union[dai.ImgDetections, dai.SpatialImgDetections, ImgDetectionsExtended],
+        ] = {}
+        self._ready_timestamps = PriorityQueue()
+
+        self.input_recognitions = dai.Node.Input(self)
+        self.input_detections = dai.Node.Input(self)
+        self.output = dai.Node.Output(self)
+
+    def build(self) -> "DetectionsRecognitionsSync":
```
jkbmrz, line 29: Should we force that the `_camera_fps` is set within the `build()` method? Because if the user doesn't remember to set it, the default value can result in wrong calculations later on.

depthai_nodes/node/detections_recognitions_sync.py
```diff
@@ -0,0 +1,161 @@
...
+        return self
+
+    def set_camera_fps(self, fps: int) -> None:
+        self._camera_fps = fps
+
+    def run(self) -> None:
+        while self.isRunning():
+            input_recognitions = self.input_recognitions.tryGet()
```
jkbmrz, line 37: should the variable be renamed from `input_recognitions` to `input_recognition` as we are drawing only one, not multiple? And same for the `input_detections` variable a few lines below.

depthai_nodes/node/detections_recognitions_sync.py
jkbmrz, line 10: Docstrings are missing.

dominik737: Yes. The `dai.node.HostNode` contains sync logic. Purpose of this node is to have different sync logic.

kkeroo:
Can we come up with a better name? Because the node can sync any other message not only recognitions?

kkeroo:
I meant that this node can also sync detections with `Keypoints` msg. for example, right?

dominik737:
Yes, technically if you have keypoints in `ImgDetectionsExtended`. Altought I would argue that it's because the `ImgDetectionsExtended` is overused and contains many other things not just detections as it's name falsely suggest.

Nonetheless better name ideas are welcome.

dominik737:
Actually, you are right. You can supply other objects beside the `NNData`. We just assume and typehint for that. 

If we want to make it more generic we should change the typehints and then the name should follow.

depthai_nodes/node/detections_recognitions_sync.py
HonzaCuhel: Fixed

depthai_nodes/node/detections_recognitions_sync.py
HonzaCuhel: Enforced

depthai_nodes/node/two_stage_sync.py
```diff
@@ -0,0 +1,198 @@
...
+            if self._timestamps_in_tolerance(detection_ts, recognition_ts):
+                self._add_recognition_by_detection_ts(recognition, detection_ts)
+                matched_recognitions.append(recognition)
+
+        for matched_recognition in matched_recognitions:
+            self._unmatched_recognitions.remove(matched_recognition)
+
+    def _timestamps_in_tolerance(self, timestamp1: float, timestamp2: float) -> bool:
+        difference = abs(timestamp1 - timestamp2)
+        return difference < (1 / self._camera_fps / self.FPS_TOLERANCE_DIVISOR)
```
jkbmrz, line 132: Should we allow the option to modify the `self.FPS_TOLERANCE_DIVISOR` param? Because, if understanding correctly, the delay depends on the "recognition" model which might be lower/higher depending on the specific architecture used.

depthai_nodes/node/two_stage_sync.py
```diff
@@ -0,0 +1,198 @@
...
+        self, recognition: dai.Buffer, detection_ts: float
+    ) -> None:
+        if detection_ts in self._recognitions_by_detection_ts:
+            self._recognitions_by_detection_ts[detection_ts].append(recognition)
+        else:
+            self._recognitions_by_detection_ts[detection_ts] = [recognition]
+
+    def _update_ready_timestamps(self, timestamp: float) -> None:
+        if not self._timestamp_ready(timestamp):
+            return
```
jkbmrz, line 144: Why is this `return` needed?

depthai_nodes/node/two_stage_sync.py
```diff
@@ -0,0 +1,198 @@
...
+    def _clear_unmatched_recognitions(self, current_timestamp) -> None:
+        unmatched_recognitions_to_remove = []
+        for unmatched_recognition in self._unmatched_recognitions:
+            if self._get_total_seconds_ts(unmatched_recognition) < current_timestamp:
+                unmatched_recognitions_to_remove.append(unmatched_recognition)
+
+        for unmatched_recognition in unmatched_recognitions_to_remove:
+            self._unmatched_recognitions.remove(unmatched_recognition)
+
+    def _get_total_seconds_ts(self, buffer_like: dai.Buffer) -> float:
```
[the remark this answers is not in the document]
jkbmrz: Also, why the `.total_seconds()` is needed. Cannot we simply compare the timestamps?

jkbmrz (changes requested):

aljazkonec1 (commented):
Added some small comments about the structure that dont affect the logic. 

The logic is sound, so LGTM

depthai_nodes/node/two_stage_sync.py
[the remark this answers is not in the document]
dominik737: The `getTimestamp` gives you `datetime.timedelta` object. With the object you cannot really easily do arithmetics like substracting plain number. For that reason the `.total_seconds()` is used instead.

When you reference something that is repeating (e.g. multiple method calls) you should create a method, otherwise you violate DRY principle. The `_get_total_seconds_ts` is referenced several times in the code. If the way to obtain total seconds would change somehow the replacement would have to be made in all cases. Also you would have to write these two methods every time you want to get the _timestamp total seconds_.

depthai_nodes/node/two_stage_sync.py
jkbmrz: Wouldn't it be much simpler (and also had same efficiency) if written as:
```
def _update_ready_timestamps(self, timestamp: float) -> None:
    if self._timestamp_ready(timestamp):
        self._ready_timestamps.put(timestamp)
```

dominik737: Efficiency is the same and negligible in this case, let's put that aside.

Simplicity is quite subjective. The one who is used to guard clauses, looks at the method and sees there are clauses that prevent further execution of the method. The method logic is less horizontally nested with the clause, which increases the readability.

If in the future there are more conditions why the ready timestamps should not be updated, it can be solved by simply adding another guard clause. Another guard clause will not make the condition more complicated and won't claim more horizontal space, this makes it quite extensible.

dominik737:
Closing the PR with successor https://github.com/luxonis/depthai-nodes/pull/203/.

PR #196, opened 2025-04-03
fix: ApplyColormap output message

jkbmrz:
## Purpose
Fixing bugs in `ApplyColormap` node:

- The `timestamp` and `sequence number` info of the outgoing message are not set correctly which is causing syncing issues in examples where the outgoing message is overlaid on top of other messages (e.g. in the `depthai-experiments/.../generic-example` with the `--overlay` attribute);
- the `process()` method is not accepting the `SegmentationMask` messages produced by some models.

## Specification

- The `timestamp` and `sequence number` values of the outgoing message are matched with those of the incoming message;
- the `process()` method is modified to also accept the `SegmentationMask` message type.

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
The changes were tested by running the updated `generic-example` (taken from [#631](https://github.com/luxonis/depthai-experiments/pull/631)) both on RVC2 and RVC4 using the following models:

- `yolov8-instance-segmentation-nano:coco-512x288`
- `midas-v2-1:small-256x192`
- `fastsam-s:512x288`
- `mediapipe-selfie-segmentation:256x144`

### DIFF
--- a/depthai_nodes/node/apply_colormap.py
+++ b/depthai_nodes/node/apply_colormap.py
@@ -2,7 +2,7 @@ import cv2
 import depthai as dai
 import numpy as np
 
-from depthai_nodes.message import ImgDetectionsExtended, Map2D
+from depthai_nodes.message import ImgDetectionsExtended, Map2D, SegmentationMask
 
 
 class ApplyColormap(dai.node.HostNode):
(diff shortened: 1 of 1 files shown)


klemen1999 (approved):
LGTM

PR #199, opened 2025-04-04
Feat: Add copy mechanism to native messages

jkbmrz:
## Purpose
Extending the copy mechanism for `depthai`/`depthai-nodes` messages.

## Specification
Adding `.copy()` methods to `depthai-nodes` messages and expanding the `copy_message` function for `depthai` messages.

None

## Deployment Plan
None

## Testing & Validation
Unittests.


jkbmrz:
@moratom, depthai offers no native option for message copying so we've started implementing this in `depthai-nodes`. Could you please check the ([_copy](https://github.com/luxonis/depthai-nodes/blob/9cf398a540eeba5e512e12430c8195ba9710aaaf/depthai_nodes/node/utils/copy_message.py#L33)) function to make sure our message copying logic is sound?

kkeroo (approved):

moratom (commented):
It will be much cleaner/future proofed to have this implemented in DepthAI directly (we don't have to hardcode any fields). 

Is having this in by the end of the month acceptable?

moratom:
@jkbmrz by the way - the default implementations of copy/deep_copy don't work as expected on your custom derived types?

jkbmrz:
> @jkbmrz by the way - the default implementations of copy/deep_copy don't work as expected on your custom derived types?

nope, there is a problem with serialization of custom Python classes.

depthai_nodes/node/utils/copy_message.py
```diff
@@ -76,12 +76,21 @@ def _copy_img_detections(
         ]
         return img_dets_copy
 
+    def _copy_point2f(point2f: dai.Point2f) -> dai.Point2f:
+        point2f_copy = _copy_metadata(point2f)
+        point2f_copy.x = point2f.x
+        point2f_copy.y = point2f.y
+        # TODO: set the value for .isNormalized()
```
klemen1999, line 83: I suppose this is still valid todo? Anything blocking it?

depthai_nodes/message/lines.py
```diff
@@ -25,6 +26,23 @@ def __init__(self):
...
+
+        @return: A new instance of the Line class.
+        @rtype: Line
+        """
+        new_obj = Line()
+        from depthai_nodes.node.utils import (
+            copy_message,
+        )
+
+        # TODO: move on top after debugging issues with circular imports
```
klemen1999, line 40: Can we handle this? Which parts are clashing?

depthai_nodes/node/utils/copy_message.py
jkbmrz: There is no way to set this manually as far as I'm concerned. cc @moratom

depthai_nodes/message/lines.py
jkbmrz: I couldn't find a fix for it but I'll give it a second look.

PR #202, opened 2025-04-10
Feat: Add BaseHostNode

jkbmrz:
## Purpose
Based on the [comment](https://github.com/luxonis/depthai-nodes/pull/192#discussion_r2021554304) in #192, we are missing an abstract class with platform extraction functionality. Adding it limits code repetition in host nodes and centralizes the control over the `ImgFrame.Type` settings for individual platforms.

## Specification
Adding the `BaseHostNode`, an abstract class with platform extraction functionality, and using it as a parent in all of our host nodes.

None

## Deployment Plan
None

## Testing & Validation
Tested by running the `depthai-experiments/neural-networks/generic-example` using the `midas-v2-1:small-256x192` model both on RVC2 and RVC4.

### DIFF
--- a/depthai_nodes/node/img_frame_sender.py
+++ b/depthai_nodes/node/img_frame_sender.py
@@ -1,0 +1,38 @@
+import depthai as dai
+from abc import ABCMeta
+
+
+HostNodeMeta = type(dai.node.HostNode)  # metaclass of dai.node.HostNode
+
+
+class CombinedMeta(HostNodeMeta, ABCMeta):
+    pass
+
+
+class ImgFrameSender(dai.node.HostNode, metaclass=CombinedMeta):
+    """An abstract base class for host nodes that send out dai.ImgFrame objects. Designed to encapsulate and abstract the configuration of platform-specific attributes,
+    providing a clean and consistent interface for derived classes.
+
+    """
+
+    IMG_FRAME_TYPES = {
+        "RVC2": dai.ImgFrame.Type.BGR888p,
+        "RVC4": dai.ImgFrame.Type.BGR888i,
+    }  # TODO: extend for other platforms?
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        self._platform = (
+            self.getParentPipeline().getDefaultDevice().getPlatformAsString()
+        )
+
+        try:
+            self._img_frame_type = self.IMG_FRAME_TYPES[self._platform]
+        except KeyError:
+            raise ValueError(
+                f"No dai.ImgFrame.Type defined for platform {self._platform}."
+            )
+
+    def process(self) -> None:
+        pass
--- a/depthai_nodes/node/img_frame_sender.py
+++ b/depthai_nodes/node/img_frame_sender.py
@@ -5,7 +5,7 @@ import depthai as dai
 HostNodeMeta = type(dai.node.HostNode)  # metaclass of dai.node.HostNode
 
 
-class CombinedMeta(HostNodeMeta, ABCMeta):
+class CombinedMeta(ABCMeta, HostNodeMeta):
     pass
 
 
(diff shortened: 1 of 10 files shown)


depthai_nodes/node/img_frame_sender.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ImgFrameSender(dai.node.HostNode, metaclass=CombinedMeta):
+    """An abstract base class for host nodes that send out dai.ImgFrame objects. Designed to encapsulate and abstract the configuration of platform-specific attributes,
+    providing a clean and consistent interface for derived classes.
+
+    """
+
+    IMG_FRAME_TYPES = {
+        "RVC2": dai.ImgFrame.Type.BGR888p,
```
klemen1999, line 19: Might be better to work with actual type rather than its string representation in case it, for whatever reason, changes down the line. So using `dai.Platform` and `.getPlatform()`

jkbmrz: Added in [ed001b3](https://github.com/luxonis/depthai-nodes/pull/202/commits/ed001b3a034ae601a56931359e7e2445f3ef3f4d).

dominik737: I think this was meant to use e.g. `dai.Platform.RVC2`.

dominik737: Also you can safely remove the `TODO` comment. Unlikely there will be any more platforms soon. And it's likely that if there are it will be not be supported in DepthAIv3.

depthai_nodes/node/img_frame_sender.py
dominik737, line 1: I would go as far as having a `outSend(cvFrame: np.NDArray)` method.

jkbmrz: Not sure if this would be of much help at this point as we must also correctly set the `timestamp` and `sequenceNum` (based on the incoming message) in both ApplyColormap and ImgFrameOverlay to get the right visualization.

dominik737: We could add also e.g. `frame1` param. But you are right it's not as nice as I initially thought.

jkbmrz: I'd leave it as it is for now but keep it in mind that such a method might come handy at some point.

dominik737 (approved):

kkeroo (approved):
LGTM, just thinking if we should rename this base class `ImageFrameSender` to more general name `BaseHostNode` or something. Because its a general class and not a "sender" and it might come useful in other nodes aswell.

klemen1999:
Maybe `BaseImgFrameSender` or `BaseFrameSender`? Because it is a base abstract class that all other nodes that send out frames could inherit from, but I wouldn't make it a BaseHostNode because that implies all host nodes inherit from it IMO

kkeroo:
Ok, but I dont see why other nodes couldnt inherit from this node. `_platform` attribute could be used in other nodes as well, similar for `_frame_type`?

jkbmrz:
> Ok, but I dont see why other nodes couldnt inherit from this node. _platform attribute could be used in other nodes as well, similar for _frame_type?

That's true. There is currently no direct usage of platform parameter in other nodes but there might come more where this might come handy. I'd thus agree we rename the node to `BaseHostNode` and use it as a parent class in all of our nodes. This might also allow us to abstract more things as we proceed with adding some new nodes.

klemen1999:
Actually I agree yeah, I think it makes sens to make it into an overall base node. Let's make sure every other inherits from it aswell

dominik737:
>  This might also allow us to abstract more things in the future.
I think having custom base node is good to abstract some very very common things. 

On the other hand I would not put too much logic that is not common/true for all our nodes. That would then over time lead to the node being very bulky, unclear and with many responsibilities as opposed to single responsibility.

klemen1999 (approved):

PR #203, opened 2025-04-11
Feat: Gather Data

dominik737:
## Purpose
Addition of `GatherData` node. The node can by default gather number of `NNData` specified by the length of `dai.ImgDetections.detections` with the same (or in tolerance) timestamp. The node can be used a n-stage sync.

## Specification

The `GatherData` node and the related tests are added. The `getAll` method was added to the mock `Queue`. The mock `Queue.send` was changed such that the Infinite queues work as FIFO not LIFO. This makes the items in the Queue being cycled, instead of the last inserted being always returned.

The change of the mock `Queue` to FIFO could change behavior of dependent tests.

## Deployment Plan
None / not applicable

## Testing & Validation
The tests for `GatherData` were made and are passing.


tests/unittests/test_nodes/test_threaded_host_nodes/test_gather_data.py
```diff
@@ -0,0 +1,236 @@
...
+    reference_timestamp,
+    tolerance,
+    duration,
+):
+    gather_data = setup_gather_data_node(
+        generator=gather_data_generator, fps=fps, duration=duration
+    )
+
+    data_items = [nn_data_in_tolerance, nn_data_in_tolerance, nn_data_out_of_tolerance]
+    results = send_data_and_run(
```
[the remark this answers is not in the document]
kkeroo: + lets add `_node` to the filename (`test_gather_data_node.py`) so stability tests can pick it up.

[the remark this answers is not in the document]
dominik737: > I see a potential problem in this approach where we first run the node and collect all the messages and when the node stops we check the msgs. If we run the tests with duration of 7 days we will get OOM. So its best to check messages on-the-fly.

That's right if you ran it for 7 days you could get OOM. The problem I had with the approach that was there is it was really dirty. The test structure should be BUILD-OPERATE-CHECK, or should at least resemble of that. If you weasel in some lambda with validation it's extremely hard to read and very unusual that inside operate you also perform check.

Why do we run these tests with `InfinityQueues` for `x` amount of time? I don't think anything could change to the logic overtime. We mock all the device components. We will not catch any device instability. Those are just plain unit tests of our Python code.

depthai_nodes/message/gathered_data.py
```diff
@@ -0,0 +1,67 @@
...
+        self.gathered = gathered
+
+    @property
+    def reference_data(self) -> TReference:
+        """Returns the reference data.
+
+        @return: Reference data.
+        @rtype: TReference
+        """
+        return self._reference_data
```
jkbmrz, line 34: Should this be `self.reference_data` (without the "_" prefix), as defined in the `__init__`?

depthai_nodes/message/gathered_data.py
```diff
@@ -0,0 +1,67 @@
...
+        self._reference_data = value
+
+    @property
+    def gathered(self) -> List[TGathered]:
+        """Returns the collected data.
+
+        @return: List of collected data.
+        @rtype: List[TGathered]
+        """
+        return self._gathered
```
jkbmrz, line 55: same as the comment above

depthai_nodes/message/gathered_data.py
```diff
@@ -0,0 +1,67 @@
+from typing import Generic, List, TypeVar
+
+import depthai as dai
+
+TReference = TypeVar("TReference", bound=dai.Buffer)
+TGathered = TypeVar("TGathered")
+
+
+class GatheredData(dai.Buffer, Generic[TReference, TGathered]):
+    """A class for gathered number of data and the reference data on which the data was
```
aljazkonec1, line 10: Could we rephrase this a bit? Its not informative when you first read it. Could just add aditional information like: "It gathers N messages based on a reference message" or something similar.

tests/unittests/test_nodes/test_threaded_host_nodes/test_gather_data.py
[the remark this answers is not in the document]
kkeroo: > Why is it necessary to name the file with suffix node? Its not a good name for the test file. The test file is in folder test_threaded_host_nodes. It's really evident that inside are tests for threaded host nodes.

Because we have also test files for some classes that are not nodes, e.g `host_spatial_calc`, and we dont need to run them in stability tests. We want to leave them out to not spawn another 7-days-long test on the testbed.

> Why do we run these tests with InfinityQueues for x amount of time? I don't think anything could change to the logic overtime. We mock all the device components. We will not catch any device instability. Those are just plain unit tests of our Python code.

Yes hopefully it doesnt, but this this is used for the stability tests where we want to assure that our nodes can run forever and not have some internal arrays, lists etc. that could fill up after long time or something else. So, with `InfinityQueue` we simulate this long running.

> We will not catch any device instability.

Its not about device instability but rather the nodes instability.

depthai_nodes/node/gather_data.py
```diff
@@ -0,0 +1,215 @@
...
+    @staticmethod
+    def _default_wait_count_fn(reference: TReference) -> int:
+        assert isinstance(reference, HasDetections)
+        return len(reference.detections)
+
+    def build(
+        self,
+        camera_fps: int,
+        wait_count_fn: Callable[[TReference], int] = _default_wait_count_fn,
+    ) -> "GatherData[TReference, TGathered]":
```
aljazkonec1, line 72: Add docstring with explanation on how and what to set for `wait_count_fn`

jkbmrz (approved):
Generally LGTM, just adding two small comments

depthai_nodes/node/gather_data.py
```diff
@@ -0,0 +1,215 @@
...
+
+TReference = TypeVar("TReference", bound=dai.Buffer)
+TGathered = TypeVar("TGathered", bound=dai.Buffer)
+
+
+class GatherData(dai.node.ThreadedHostNode, Generic[TReference, TGathered]):
+    FPS_TOLERANCE_DIVISOR = 2.0
+    INPUT_CHECKS_PER_FPS = 100
+    """A class for gathering data. By default gathers n of dai.NNData where n is number
+    of dai.ImgDetection objects in dai.ImgDetections.
```
aljazkonec1, line 34: Not necessarily dai.Imgdetection objects. I would rewrite and add more precise description of what the node does. Eg. Mention it waits for exactly n messages based on reference_data. To determine n, wait_count_fn function is used, by default set to num detections. When all n messages are collected, all messages and reference messages with a lower timestamp are dropped.

aljazkonec1 (commented):
Overall the logic is the same as in previous PR and is good. Left some small comments on the docstrings.

tests/unittests/test_nodes/test_threaded_host_nodes/test_gather_data.py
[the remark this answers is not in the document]
dominik737: > Yes hopefully it doesnt, but this this is used for the stability tests where we want to assure that our nodes can run forever and not have some internal arrays, lists etc. that could fill up after long time or something else.

Okay, that makes sense. I'll think about how could this be tackled.

depthai_nodes/message/gathered_data.py
jkbmrz: You are right, my bad. Just wanted to point out that we return here a private variable (`self._reference_data`) which isn't the same as the one defined in the `__init__` (`self.reference_data`). I guess that we want to have them the same - both private or not - right?

tests/unittests/test_nodes/test_threaded_host_nodes/test_gather_data.py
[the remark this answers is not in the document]
dominik737: I've transformed the code to use lambdas to not block this PR. But I think that the refactoring of the mechanism that the long running tests are implemented/approached should be performed.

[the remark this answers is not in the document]
dominik737: Also I've added the `_node` suffix. That's also an area for potential improvement. It's very unclear to the developer that he should add redundant suffix.

depthai_nodes/message/gathered_data.py
dominik737: Not sure whether I fully understand what you mean. There is public property and private backing variable. In the constructor I'm setting the property to use the validation that is connected to the property.

depthai_nodes/message/gathered_data.py
dominik737: Sure, I can rephrase. I'm proposing `Contains N messages and reference data that the messages were matched with.`. What do you think? Keep in mind the message is not doing any gathering, thus I'd like to avoid the `It gathers` phrasing.

kkeroo (approved):
LGTM

depthai_nodes/message/gathered_data.py
aljazkonec1: Yes sounds good

aljazkonec1 (approved):
LGTM, just the last docstring update

PR #205, opened 2025-04-17
Default AnnotationHelper colors.

kkeroo:
## Purpose
`AnnotaitonHelper` node can have default colors which are aligned with official Luxonis colors. Same goes for default visualizations of messages.

## Specification
Added primary, secondary, and primary transparent (used as fill color in bboxes - similar to Visualizer).

It has some naming changes which may lead to broken experiments.

## Deployment Plan
None / not applicable

## Testing & Validation
Tested on affected messages.

### DIFF
--- a/depthai_nodes/constants.py
+++ b/depthai_nodes/constants.py
@@ -1,6 +1,11 @@
 import depthai as dai
 
-OUTLINE_COLOR = dai.Color(1.0, 0.5, 0.5, 1.0)
-TEXT_COLOR = dai.Color(0.5, 0.5, 1.0, 1.0)
-BACKGROUND_COLOR = dai.Color(1.0, 1.0, 0.5, 1.0)
-KEYPOINT_COLOR = dai.Color(1.0, 0.35, 0.367, 1.0)
+PRIMARY_COLOR = dai.Color(
+    float(78 / 255), float(56 / 255), float(237 / 255), float(1.0)
+)
+TRANSPARENT_PRIMARY_COLOR = dai.Color(
+    float(78 / 255), float(56 / 255), float(237 / 255), float(0.2)
+)
+SECONDARY_COLOR = dai.Color(
+    float(240 / 255), float(240 / 255), float(240 / 255), float(1.0)
+)
--- a/depthai_nodes/constants.py
+++ b/depthai_nodes/constants.py
@@ -1,10 +1,10 @@
 import depthai as dai
 
 PRIMARY_COLOR = dai.Color(
-    float(78 / 255), float(56 / 255), float(237 / 255), float(1.0)
+    float(87 / 255), float(36 / 255), float(232 / 255), float(1.0)
 )
 TRANSPARENT_PRIMARY_COLOR = dai.Color(
-    float(78 / 255), float(56 / 255), float(237 / 255), float(0.2)
+    float(87 / 255), float(36 / 255), float(232 / 255), float(0.2)
 )
 SECONDARY_COLOR = dai.Color(
     float(240 / 255), float(240 / 255), float(240 / 255), float(1.0)
--- a/depthai_nodes/message/classification.py
+++ b/depthai_nodes/message/classification.py
@@ -5,10 +5,7 @@ import depthai as dai
 import numpy as np
 from numpy.typing import NDArray
 
-from depthai_nodes.constants import (
-    BACKGROUND_COLOR,
-    TEXT_COLOR,
-)
+from depthai_nodes import PRIMARY_COLOR, SECONDARY_COLOR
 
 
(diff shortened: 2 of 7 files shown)


depthai_nodes/message/classification.py
```diff
@@ -159,14 +156,22 @@ def getVisualizationMessage(self) -> dai.ImgAnnotations:
...
         img_annotations = dai.ImgAnnotations()
         annotation = dai.ImgAnnotation()
+        w, h = self.transformation.getSize()
 
-        for i in range(len(self._scores)):
+        font_size = h / 30
+        x_offset = 2 / w
+        y_offset = 2 / h
+
+        for i in range(5):
```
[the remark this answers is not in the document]
kkeroo: Fixed in [7e7cd16](https://github.com/luxonis/depthai-nodes/pull/205/commits/7e7cd16c53fffcb7f1a9d39e5407a8318ab5d8fa).

depthai_nodes/message/img_detections.py
```diff
@@ -332,6 +331,9 @@ def getVisualizationMessage(self) -> dai.ImgAnnotations:
         transformation = self.transformation
         w, h = transformation.getSize()
 
+        if w == 0 or h == 0:
```
jkbmrz, line 334: In what case `w == 0` or `h == 0`? Do we even expect there be any detections in that case?

depthai_nodes/message/img_detections.py
```diff
@@ -341,16 +343,20 @@ def getVisualizationMessage(self) -> dai.ImgAnnotations:
...
-            pointsAnnotation.outlineColor = OUTLINE_COLOR
-            pointsAnnotation.thickness = 2.0
+            pointsAnnotation.outlineColor = PRIMARY_COLOR
+            pointsAnnotation.fillColor = TRANSPARENT_PRIMARY_COLOR
+            pointsAnnotation.thickness = 1.0
             annotation.points.append(pointsAnnotation)
 
             text = dai.TextAnnotation()
             text.position = points[0]
+            font_size = h / 30
```
jkbmrz, line 353: Why do we adjust the `font_size` to `h` and not to `w` or a combination of two? Also, do we even need to adjust the `font_size`? Keeping it constant should work as the `DAI Visualizer` keeps the original dimensions of the data (so choosing a good constant value should always be readable).

depthai_nodes/constants.py
```diff
@@ -1,6 +1,11 @@
...
-KEYPOINT_COLOR = dai.Color(1.0, 0.35, 0.367, 1.0)
+PRIMARY_COLOR = dai.Color(
+    float(78 / 255), float(56 / 255), float(237 / 255), float(1.0)
+)
+TRANSPARENT_PRIMARY_COLOR = dai.Color(
+    float(78 / 255), float(56 / 255), float(237 / 255), float(0.2)
+)
+SECONDARY_COLOR = dai.Color(
+    float(240 / 255), float(240 / 255), float(240 / 255), float(1.0)
+)
```
jkbmrz, line 11: Out of curiosity, how did you determine these values?

jkbmrz (approved):
LGTM, just a few small remarks.

depthai_nodes/message/img_detections.py
jkbmrz: (This comment is also relevant to other messages where we adjust the `font_size`)

klemen1999:
> It has some naming changes which may lead to broken experiments.

Let's have a PR ready on dai-experiments side that fixes this if needed (and bump depthai-nodes version in requirements of these experiments to next dai-nodes version)

depthai_nodes/message/img_detections.py
kkeroo: Thanks, forgot to remove that. Its a hotfix to the potential bug in a15.

depthai_nodes/message/img_detections.py
kkeroo: Hm if we set the font size to constant value (which unit I believe are pixels) than the size will not adjust based on the frame size e.g. font will look normal on YOLO 512x288, but will be extremely large on palm-detection 128x128.

kkeroo:
Fix in dai-experiments in this [PR](https://github.com/luxonis/depthai-experiments/pull/652).

depthai_nodes/message/img_detections.py
jkbmrz: > font will look normal on YOLO 512x288, but will be extremely large on palm-detection 128x128

That's true, but on the other hand, adjusting it to the image size, it will also make it pretty small on the 128x128 image. But I don't have a strong opinion on this so this comment shouldn't be blocking the merge!

PR #206, opened 2025-04-17
feat: generate_script_content update

jkbmrz:
## Purpose
Platform differentiation in the `generate_script_content` function is not necessary anymore because we can now utilize `ImageManipConfigV2` both for RVC2 and RVC4.

## Specification
- update `generate_script_content` function:
    - remove platform differentiation during `cfg_content` creation,
    - add `resize_mode` setting;
- update associated unittests in the `unittests/test_nodes/test_host_nodes/test_detection_config_generator.py` na rename it to `test_generate_script_content.py`

Removing the `platform` parameter from the `generate_script_content` function will cause some disturbances. Thankfully, none of the examples in `depthai-experiments` imports this function (but they rather use custom implementations). 
    
## Deployment Plan
None

## Testing & Validation
Running the updated `human-pose` experiment ([HERE](https://github.com/luxonis/depthai-experiments/pull/650)) on RVC2.


kkeroo (approved):
LGTM.

jkbmrz:
The `test_generate_script_content.py` is named a bit awkwardly. Could we rename it to `test_generate_script_content.py`? This seems more clear and also follows the naming convention we use elsewhere (`test_<class/function_name>`)

depthai_nodes/node/utils/detection_config_generator.py
```diff
@@ -49,11 +45,9 @@ def generate_script_content(
             rect.size.height = rect.size.height + {padding} * 2
             rect.angle = 0
 
-            cfg.addCropRotatedRect(rect=rect, normalizedCoords=True)
-            cfg.setOutputSize({resize_width}, {resize_height})
+            cfg.addCropRotatedRect(rect, True)
```
klemen1999, line 48: Nitpick: Let's add `normalizedCoords` back as keyword for better readability
```suggestion
            cfg.addCropRotatedRect(rect, normalizedCoords=True)
```

klemen1999 (approved):
LGTM

PR #208, opened 2025-04-24
feat: ParsingNeuralNetwork add property to send out outputs of multiple heads in a single message

jkbmrz:
## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/depthai_nodes/node/parsing_neural_network.py
+++ b/depthai_nodes/node/parsing_neural_network.py
@@ -60,6 +60,11 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         return list(self._parsers.values())[0].out
 
     @property
+    def outputs(self) -> dai.Node.Output:
+        """Returns outputs of all heads in a single output message."""
+        pass
+
+    @property
     def passthrough(self) -> dai.Node.Output:
         return self._nn.passthrough
 
--- a/depthai_nodes/node/parsing_neural_network.py
+++ b/depthai_nodes/node/parsing_neural_network.py
@@ -23,6 +23,8 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         Neural network inputs.
     out : Node.Output
         Neural network output. Can be used only when there is exactly one model head. Otherwise, getOutput method must be used.
+    outputs: Node.Output
+        Neural network output having dai.MessageGroup as a payload which contains outputs of all model heads and can be accessed as a dictionary with str(model head index) as a key. Can be used only when there is at least two model heads. Otherwise, out property must be used.
     passthrough : Node.Output
         Neural network passthrough.
     passthroughs : Node.OutputMap
@@ -40,6 +42,7 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         self._pipeline = self.getParentPipeline()
         self._nn = self._pipeline.create(dai.node.NeuralNetwork)
         self._parsers: Dict[int, BaseParser] = {}
+        self._internal_sync: Optional[dai.node.Sync] = None
 
     @property
     def input(self) -> dai.Node.Input:
@@ -62,7 +65,11 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
     @property
     def outputs(self) -> dai.Node.Output:
         """Returns outputs of all heads in a single output message."""
-        pass
+        if self._internal_sync is None:
+            raise RuntimeError(
+                "ParsingNeuralNetwork node must have at least two model heads to use sync node and outputs property."
+            )
+        return self._internal_sync.out
 
     @property
     def passthrough(self) -> dai.Node.Output:
@@ -260,6 +267,10 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         self._nn.build(input, self._nn_archive, **kwargs)
 
         self._updateParsers(self._nn_archive)
+
+        if len(self._parsers) > 1:
+            self._createSyncNode()
+
         return self
 
     def run(self) -> None:
@@ -268,6 +279,22 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
         Method runs with start of the pipeline.
         """
         self._checkNNArchive()
+
+    def cleanup(self):
+        """Cleans up the ParsingNeuralNetwork node and removes all nodes created by
+        ParsingNeuralNetwork from the pipeline.
+
+        Must be called before removing the node from the pipeline.
+        """
+        if self._internal_sync is not None:
+            self._pipeline.remove(self._internal_sync)
+            self._internal_sync = None
+
+        self._pipeline.remove(self._nn)
+        self._nn = None
+
+        self._removeOldParserNodes()
+        self._parsers = {}
 
     def _updateParsers(self, nnArchive: dai.NNArchive) -> None:
         self._removeOldParserNodes()
@@ -312,3 +339,14 @@ class ParsingNeuralNetwork(dai.node.ThreadedHostNode):
 
     def _getConfig(self):
         return self._nn_archive.getConfig()
+
+    def _createSyncNode(self):
+        if len(list(self._parsers.values())) <= 1:
+            raise RuntimeError(
+                "ParsingNeuralNetwork node must have at least two model heads to use sync node."
+            )
+        self._internal_sync = self._pipeline.create(dai.node.Sync)
+        self._internal_sync.setRunOnHost(True)
+        outputs = [parser.out for parser in list(self._parsers.values())]
+        for ix, output in enumerate(outputs):
+            output.link(self._internal_sync.inputs[str(ix)])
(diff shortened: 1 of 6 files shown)


jkbmrz:
I'm a bit stuck on this. Within `ParsingNeuralNetwork` we operate with parser objects and we can only get `dai.Node.Output` out of them (and not the actual messages). I don't really see a way how to reconstruct these into `dai.MessageGroup` without an additional host node. Based on our discussion [HERE](https://github.com/luxonis/depthai-experiments/pull/650#discussion_r2056425156), do you have some ideas how to tackle this problem? @dominik737 @kkeroo @klemen1999

kkeroo:
Hm yeah, looking into it its not that trivial to implement. One idea how we could have it done is by utilizing `Sync` node which is silently added to the pipeline if the `ParsingNeuralNetwork` detects multihead model. How it could work is that we create `Sync` node inside PNN with the `parentPipeline` during the build if its multihead model.

```
        self.mux = self._pipeline.create(dai.node.Sync)
        self.mux.setRunOnHost(True)
        outputs = [parser.out for parser in list(self._parsers.values())]
        for ix, output in enumerate(outputs):
            output.link(self.mux.inputs[str(ix)])
```

And the `outputs` property can then be defined as 
```
    @property
    def outputs(self) -> dai.Node.Output:
        return self.mux.out
```

The output msg will then be `MessageGroup` and I think syncing wont be problem because all heads will have the same timestamp etc.

Alternatively, we can just use `Sync` node itself in the experiments pipeline to group the outputs before going into the `GatherData`. Or we create new host node which just groups the incoming messages (`Group`/`Multiplexer` etc)

jkbmrz:
Augmenting the `ParsingNeuralNetwork` with `Sync` node would be quite a nice solution. However, using the `Sync` node from `dai` might cause latency due to host/device switching (for example, the information flow of the `age-gender` experiment would be as follows: ... -> parsers run on host -> `dai.node.Sync` runs on device -> `depthai_nodes.node.GatherNode` runs on host). Thus, making a custom version of the `Sync` that runs on host might be the best performant solution to this problem (and we can still include it silently to the `ParsingNeuralNetwork`).

kkeroo:
> However, using the Sync node from dai might cause latency due to host/device switching

`Sync` node can also run on host (`self.mux.setRunOnHost(True)`).

dominik737:
You can either add the `Sync` node, but need to ensure for example that when the `ParsingNeuralNetwork` node gets removed from the pipeline also the `Sync` node gets removed from the pipeline. 

Or you can have some very simple internal mechanism for this. You can create a queue on the `dai.Node.Output` with `createOutputQueue()`, you set it `blocking=True`, since you don't want to throw the messages. 
Then you can either have:
1. A cycle in another thread that would call `get` on all the head output queues and then would output a `dai.MessageGroup` out of it. 
2. On the output queue call `addCallback` and call a function that would append the item to thread-safe dictionary. After the append it would check whether all head outputs are present. If so it would output the `dai.MessageGroup`.

jkbmrz:
Thanks! I think adding the `Sync` node would be a more clean solution (also the code will be more readable).

kkeroo:
In [72096ef](https://github.com/luxonis/depthai-nodes/pull/208/commits/72096efaaf0f2cdae4d2790cfdeae16202eb359f) we implement `outputs` property in PNN which uses internal `Sync` node. PNN also has `cleanup` function which should be called before `pipeline.remove()` method - we clean up all nodes created by PNN. Currently, there is no option to replace default remove method in pipeline because it is read-only. Nodes also dont have some callback that is triggered when the node is deleted so manual cleanup must be done.

Additionally, deafult resize mode is set to `STRETCH` because `NONE` does not guarantees that object will be fully visible.

jkbmrz (commented):
Looking great! Thanks for adding the changes @kkeroo

klemen1999 (approved):
LGTM

depthai_nodes/node/parsing_neural_network.py
```diff
@@ -55,10 +58,19 @@ def out(self) -> dai.Node.Output:
...
             raise RuntimeError(
                 f"Property out is only available when there is exactly one model head. \
-                               The model has {self._getModelHeadsLen()} heads. Use {self.getOutput.__name__} method instead."
+                               The model has {self._getModelHeadsLen()} heads. Use {self.getOutput.__name__} method or {self.outputs.__name__} property instead."
             )
         return list(self._parsers.values())[0].out
 
+    @property
+    def outputs(self) -> dai.Node.Output:
+        """Returns outputs of all heads in a single output message."""
```
dominik737, line 67: Maybe I would mention here as well that the `dai.Node.Output` returns `dai.MessageGroup` messages.

tests/utils/nodes/mocks/sync_node.py
dominik737, line 1: Do we need to mock this node in case when it runs on host?

depthai_nodes/node/parsing_neural_network.py
```diff
@@ -264,6 +280,22 @@ def run(self) -> None:
...
+        """
+        if self._internal_sync is not None:
+            self._pipeline.remove(self._internal_sync)
+            self._internal_sync = None
+
+        self._pipeline.remove(self._nn)
+        self._nn = None
+
+        self._removeOldParserNodes()
+        self._parsers = {}
```
dominik737, line 297: @moratom Is there any way we could call this method when `pipeline.remove` is called on this node? 

I see potentially there is `createSubnode` function as a part of each node API, do these get deleted automatically with the parent node? Also what is the signature? The stub has only this `createSubnode: ClassVar[Callable] = ...` from which it's unclear what parameters the function takes.

moratom: @dominik737 not right now, except maybe for some python internal function that get's called on destruction (`__del__`) but that generally shouldn't be overriden.

`createSubnode` takes a class type, that has to be a node.

dominik737: I think destructor won't work in this case, since you can have the node reference while removing the node from pipeline.

> `createSubnode` takes a class type, that has to be a node.

So can we create the subnodes using this function and they will get removed when the parent node that created the subnodes gets removed?

kkeroo: > Out of curiousity, will depthai-nodes actually be using the feature to remove nodes from the pipeline?
It was added awhile back for a "depthai-wrapper" like project, but it's not rutinely used, so there might be caviats.

Currently, we are not removing nodes inside dai-nodes and honestly I dont think we will in the future. So might be worth to wait for the use-case before handling the nodes removal? @dominik737

dominik737: I think most nodes in this repository are not composed of other existing nodes. Exception to this is the `ParsingNeuralNetwork`. The `remove` method is used e.g. [here](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsing_neural_network.py#L271). When the `setNNArchive` is called the parser nodes have to be updated - meaning the old ones have to get removed.

dominik737: @jkbmrz With those new informations, could we remove the `cleanup` method and use the `self.createSubnode` instead of `pipeline.create`? This applies even for the NN node.

jkbmrz: > use the `self.createSubnode` instead of `pipeline.create`? This applies even for the NN node.

Sounds good for the Sync node, but for others I'm not sure I understand how this would be done (we have 3 `pipeline.create` calls within the PNN). What would be the parent node in this case?

kkeroo: Parent node would be PNN, and you create parsers and NN (and also Sync) with self.createSubnode() and it should remove all "3" subnodes once we call pipeline.remove(pnn), but it does not sadly.

dominik737: The NN node initiated [here](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsing_neural_network.py#L41) can be child to the `ParsingNeuralNetwork` - just do `self.createSubnode` in the `ParsingNeuralNetwork`.

Then there is [the parser generator creation](https://github.com/luxonis/depthai-nodes/blob/main/depthai_nodes/node/parsing_neural_network.py#L276), which is a bit tricky when I think about it. 
We want the parsers to be also deleted when the `ParsingNeuralNetwork` is removed. Thus they should also be created using the `createSubnode`, but the parent node needs to be the `ParsingNeuralNetwork` not `ParserGenerator`, which is deleted right away after generating the parsers. I think that can be addressed in a separate PR.

Last `pipeline.create` reference should be the one to `Sync` node. And there the same applies as for the NN node.

kkeroo: Deleting the children nodes doesnt work when using `self.createSubnode()`, only the PNN gets deleted. I added proper deletion of Sync node when we set new NN archive and will leave `cleanup` for now until we figure out something better. Changes in [a7978ac](https://github.com/luxonis/depthai-nodes/pull/208/commits/a7978ac722b75b456f4b05bb4c8dad2464d816f4). We can address this in new PR.

PR #212, opened 2025-05-08
Tests: Stability tests update & better E2E logging. 

kkeroo:
## Purpose
We need better logging in E2E tests so we can see where the tests stopped. In the stability tests we need intermediate logs on all tests, leave out some unnecessary tests for stability (pnn, parser_generator), fix duration bug in some tests, add better description and improve memory limits on docker containers.

## Specification
None / not applicable

None / not applicable

## Deployment Plan
None / not applicable

## Testing & Validation
Tested with HIL team.

### DIFF
--- a/tests/end_to_end/manual.py
+++ b/tests/end_to_end/manual.py
@@ -23,14 +23,15 @@ if not (args.nn_archive or args.model):
 
 try:
     device = dai.Device(dai.DeviceInfo(args.ip))
+    print(f"(1) Connected to device with IP/mxid: {args.ip}")
 except Exception as e:
     print(e)
-    print("Can't connect to the device with IP/mxid: ", args.ip)
+    print("Can't connect to the device with IP/mxid: %s", args.ip)
     exit(6)
 
 with dai.Pipeline(device) as pipeline:
     camera_node = pipeline.create(dai.node.Camera).build()
-
+    print("(2) Camera node created.")
     if args.model:
         model_desc = dai.NNModelDescription(
             model=args.model,
@@ -38,6 +39,7 @@ with dai.Pipeline(device) as pipeline:
         )
         try:
             nn_archive_path = dai.getModelFromZoo(model_desc, useCached=False)
+            print("(3) NN archive downloaded.")
         except Exception as e:
             print(e)
             print(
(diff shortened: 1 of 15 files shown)


tests/unittests/test_nodes/test_host_nodes/test_apply_colormap_node.py
```diff
@@ -87,7 +89,13 @@ def test_processing(
         create_img_detections_extended(masks=ARR),  # ImgDetectionsExtended
     ]:
         start_time = time.time()
+        last_log_time = time.time()
         while time.time() - start_time < duration:
+            if time.time() - last_log_time > LOG_INTERVAL:
+                print(
+                    f"Test running... {time.time()-start_time:.1f}s elapsed, {duration-time.time()+start_time:.1f}s remaining"
+                )
+                last_log_time = time.time()
```
jkbmrz, line 98: Nitpicking, but `last_log_time = time.time()` could be moved to line 94 and then you don't need to define it in both line 92 and 98.

tests/unittests/test_nodes/test_host_nodes/test_apply_colormap_node.py
```diff
@@ -69,9 +73,7 @@ def test_parameter_setting(
 @pytest.mark.parametrize(
     "colormap_value", [cv2.COLORMAP_HOT, cv2.COLORMAP_PLASMA, cv2.COLORMAP_INFERNO]
 )
-def test_processing(
-    colorizer: ApplyColormap, colormap_value: int, duration: int = 1e-6
-):
+def test_processing(colorizer: ApplyColormap, colormap_value: int, duration: int):
```
jkbmrz, line 76: We define `duration: int` but this is not always true because above we define that duration can also be `1e-6` which is `float`.

jkbmrz (approved):
LGTM, just adding two small remarks.

tests/end_to_end/manual.py
```diff
@@ -23,21 +23,23 @@
 
 try:
     device = dai.Device(dai.DeviceInfo(args.ip))
+    print(f"(1) Connected to device with IP/mxid: {args.ip}")
```
klemen1999, line 26: Should we rather use logger (also inside all other tests) than print? That seems like a better longterm solution

klemen1999 (approved):
LGTM, just one suggestion

tstrus97 (approved):
LGTM from hil side

tests/unittests/test_nodes/test_host_nodes/test_apply_colormap_node.py
kkeroo: Hmm, not sure if possible because `last_log_time` stores the time when we printed out the last time.

PR #214, opened 2025-05-20
Feat/snaps producer

klemen1999:
## Purpose

Added `SnapsProducerFrameOnly` (LMK if you have a better name suggestion) and `SnapsProducer` nodes that abstract sending snaps to the backend. It allows either out of the box sending of frames depending on the time interval (default) or making custom processing functions so it can be extended for particular use case.

Example usage with custom processing function:
```python
def filter_detections_frame_only(producer, frame):
        producer.sendSnap(name="test", frame=frame)

snaps_producer = pipeline.create(SnapsProducerFrameOnly).build(
        frame=nn_with_parser.passthrough,
        process_fn=filter_detections_frame_only,
        time_interval=5,
)
```

Example with `partial` to pass in extra parameters:
```python
def partial_process_fn(
    producer: SnapsProducerFrameOnly, frame: dai.ImgFrame, threshold: float
):
    if frame.getWidth() > threshold:
        producer.sendSnap("threshold_snap", frame)

snaps_producer = pipeline.create(SnapsProducerFrameOnly).build(
        frame=nn_with_parser.passthrough,
        process_fn=partial(partial_process_fn, threshold=200)
)
```

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -25,7 +25,7 @@ on:
       depthai-version:
         description: 'Version of depthai to install. Default: alpha13'
         required: true
-        default: '3.0.0a14'
+        default: '3.0.0a15'
   push:
     branches:
       - main
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -32,7 +32,7 @@ jobs:
         cache: pip
 
     - name: Install depthai
-      run: pip install --extra-index-url ${{ secrets.LUXONIS_EXTRA_INDEX_URL }} depthai==3.0.0a14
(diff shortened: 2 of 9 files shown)


klemen1999:
Bumped DAI version to `3.0.0a15` because `a14` was problematic in terms of snaps

kkeroo (approved):
LGTM, two nodes are because there is no option to pass optional output to the process function?

klemen1999:
> LGTM, two nodes are because there is no option to pass optional output to the process function?
Yeah theoretically there are ways to go around this (factory node or some "hacky" Threaded host node) but we can try improving after release becase in most cases `SnapProducer` is probably going to be preferred way to do it.

PR #223, opened 2025-06-19
[FIX] Float confidence

aljazkonec1:
## Purpose

A special case happens in confidence setter when confidence is above 1 and below 1.1. It clips the confidence to min(0, max(1, conf)) which can later throw an error as 0 and 1 are not floats.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999:
Actually maybe it's better to explicitly cast what comes out of min-max normalization to `float(...)` for clarity just so we don't revert this mistake back. And maybe add a small test case for it

PR #225, opened 2025-07-01
Parser optimizations.

kkeroo:
## Purpose
Some of the parsers are not the optimal.

## Specification
- `YuNetParser` - cached anchors,  unnecessary np.where (replaced with np.clip), general vectorization
- `YOLOExtendedParser` - decoding speed up and pre-filtering before the NMS
- `PPDetParser` - 1. Dilation kernel reduced: Faster, less small blobs, 2. Small areas filtered before sorting: Reduces number of contours to sort/loop. 3. Top-N selection uses np.argpartition: O(N) not O(N log N) as with full sort. 4. Direct construction of output arrays: Minimizes list->array conversion overhead. 5. Less shape shuffling, no unnecessary np.reshape.
- `MLSDParser` - Explicit python loop removed and replaced with numpy; usage of batched gather
- `SCRFDParser` - Caching anchors
- `MPPalmDetectionParser` - Caching anchor
- `FastSAMParser` - yolo part decoding speed up

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/depthai_nodes/node/parsers/scrfd.py
+++ b/depthai_nodes/node/parsers/scrfd.py
@@ -72,6 +72,9 @@ class SCRFDParser(DetectionParser):
         self.num_anchors = num_anchors
         self.input_size = input_size
         self.label_names = ["Face"]
+        self._cached_anchors = self.compute_anchor_centers(
+            self.feat_stride_fpn, self.input_size, self.num_anchors
+        )
         self._logger.debug(
             f"SCRFDParser initialized with output_layer_names={output_layer_names}, conf_threshold={conf_threshold}, iou_threshold={iou_threshold}, max_det={max_det}, input_size={input_size}, feat_stride_fpn={feat_stride_fpn}, num_anchors={num_anchors}"
         )
@@ -155,6 +158,9 @@ class SCRFDParser(DetectionParser):
         self.output_layer_names = output_layers
         self.feat_stride_fpn = head_config.get("feat_stride_fpn", self.feat_stride_fpn)
         self.num_anchors = head_config.get("num_anchors", self.num_anchors)
+        self._cached_anchors = self.compute_anchor_centers(
+            self.feat_stride_fpn, self.input_size, self.num_anchors
+        )
 
         self._logger.debug(
             f"SCRFDParser built with output_layer_names={self.output_layer_names}, feat_stride_fpn={self.feat_stride_fpn}, num_anchors={self.num_anchors}"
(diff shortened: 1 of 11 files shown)


depthai_nodes/node/parsers/utils/yolo.py
```diff
@@ -407,19 +408,53 @@ def decode_yolo_output(
...
+    @param max_nms: Maximum number of boxes to keep after NMS.
+    @type max_nms: int
     @return: NMS output.
     @rtype: np.ndarray
     """
     num_outputs = num_classes + 5
-    output = parse_yolo_outputs(
-        yolo_outputs, strides, num_outputs, anchors, kpts, det_mode, subtype
-    )
+
```
klemen1999, line 417: I believe now `parse_yolo_outputs` is unused so we could remove it?

depthai_nodes/node/parsers/scrfd.py
```diff
@@ -256,3 +263,33 @@ def run(self):
             self.out.send(message)
 
             self._logger.debug("Detection message sent successfully")
+
+    def compute_anchor_centers(
```
klemen1999, line 267: This function can be moved to utils since it doesn't really need any class attributes?

klemen1999 (commented):
During "cleanup" we can remove some of the excessive comments

depthai_nodes/node/parsers/utils/yolo.py
kkeroo: Yes, removed it, actually I saw that we can you this faster decoding inside FastSAM parser (I added that also).

tersekmatija:
bugbot run

jkbmrz (approved):
LGTM if tested properly.

depthai_nodes/node/parsers/utils/yunet.py
```diff
@@ -199,3 +197,82 @@ def format_detections(
...
+    @param iou: The predicted IoU (Intersection over Union) scores.
+    @param conf_threshold: The confidence threshold for pruning.
+    @param anchors: Pre-computed anchors to avoid regeneration.
+    @param variance: A list of variances used to decode the bounding box predictions.
+    @return: A tuple of bboxes, keypoints, and scores.
+    """
+    w, h = input_size
+
+    if variance is None:
+        variance = [0.1, 0.2]
```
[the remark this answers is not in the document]
kkeroo: We cant use mutable objects (lists) like default values. This variance setting was also present before the PR. Other option is to define the variance argument in the function calling inside the parser. But it is essentially the same.

PR #229, opened 2025-07-17
Optimize E2E tests & depthai version bump.

kkeroo:
## Purpose
HIL doesnt pick up the tests long time.

## Specification
Separate RVC2 and RVC4 job, selects two testsbeds, RVC2 and RVC4 in parallel. Better testbed selection.


klemen1999 (approved):
LGTM
Which python version will it use now since we removed the explicit 3.12 from workflow?

kkeroo:
It was unnecessary, I copied the reservation name template from oak-examples back then. We use py 3.8 during the tests (see [here](https://github.com/luxonis/depthai-nodes/blob/eed933b4498a14df88a7bf6643a411ec043210bb/tests/end_to_end/Dockerfile#L2)).

PR #230, opened 2025-07-30
Feedback On Design Decisions

thehiddenwaffle:
https://discuss.luxonis.com/d/6203-are-depthai-core-contributions-welcome/9

Was informed that the team is planning this in the long term anyway so my code could be helpful. 

Quick background on me, when it comes to more explicit/imperative languages I have magnitudes more experience in Rust(and to a lesser extent C) than C++ so if some of my code comes off like I wish I was writing in Rust that's because I do wish that :sweat_smile: . Have been writing DepthAI code for about 2 years in Python so I'm very familiar with the "feel" of things from the Python side and have often used the depthai-core source as clarification for what's going on behind the scenes so I'm generally familiar with the C++ code base. 

I made some design decisions that I would love feedback on:
- Built `Keypoints` message as a template allowing for any combination of XYZ with confidence values for any/all/none of the values(SimCC provides discrete confidence values for every dimension by virtue of if being basically a heatmap made of buckets).
- Used constructor rather than a factory to create `Keypoints` message, also took ownership of input data(thinking of a C++ version of Rust's `impl TryFrom<(NNdata, xt::xarray)> for Keypoint`)
- Split BaseParser::build() into 2 pieces, build() in the parent which checks parser type and calls buildImpl() which is overridden by subclasses, allowing SimCCKeypointParser to reuse logic from it's parent implementation of buildImpl.
- Abandoned CRTP because I don't think it's worth as much for this more flexible/agile environment where we want it to be easy to subclass KeypointParser for example(very unsure if this is the right choice, again Rust/C dev who's never used CRTP).


thehiddenwaffle:
Absolutely, I was noticing a large volume of replicated cmake configuration and as you probably noticed I had to directly copy some of the internal utils so having it in core makes a lot of sense to me. I have made many improvements not listed in this commit history and it'll take me a small amount of time to get this integrated into core but I'll probably have all that completed in about 2 days.

PR #232, opened 2025-08-04
Apply objectness-based filtering only when real objectness scores are present

sokovninn:
## Purpose
In some YOLO variants the “objectness” score is not meaningful (e.g. all ones). Filtering by objectness before non-maximum suppression can therefore discard valid, high-confidence detections. This change ensures we only apply objectness-based filtering when the model actually produces different objectness scores.

## Specification

- Only if not all objectness values are identical and the total number of detections exceeds `max_nms` do we sort and trim by objectness.
- Otherwise we fall back to the previous behavior of applying NMS on all detections regardless of objectness.

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/depthai_nodes/node/parsers/utils/yolo.py
+++ b/depthai_nodes/node/parsers/utils/yolo.py
@@ -393,7 +393,7 @@ def decode_yolo_output(
 
     # 2. Concatenate all kept candidates at once
     output = np.concatenate(filtered_outputs, axis=0)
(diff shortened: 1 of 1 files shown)


sokovninn:
> LGTM. Are AI comments applicable? Lets resolve them even if not applicable

Resolved them in the second commit already

PR #238, opened 2025-08-29
Fix: Add transformations to ApplyColormap node

aljazkonec1:
## Purpose

ApplyColormap node did not propagate image transformations from its input to the output. This could cause crashes and wrong behaviour in other (depthai) nodes that rely on transformation.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable


klemen1999 (approved):
LGTM. Can we check if there are any other nodes where this is missing?

PR #239, opened 2025-08-29
FIX: Add checks if ImgTransformations is None + propagate device timestamp

aljazkonec1:
## Purpose

In certain circumstances, ImgTransformations can be set to None and would cause a crash due to the definition of setTransformation in depthai. This PR Adds if checks for cases when ImgTransformations are None.

In addition, this PR adds propagation of the deviceTimestamp via setDeviceTimestamp to all message types in depthai-nodes.
 

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

### DIFF
--- a/depthai_nodes/message/classification.py
+++ b/depthai_nodes/message/classification.py
@@ -1,5 +1,5 @@
 import copy
-from typing import List
+from typing import List, Optional
 
 import depthai as dai
 import numpy as np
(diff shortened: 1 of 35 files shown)


depthai_nodes/message/classification.py
```diff
@@ -27,7 +27,7 @@ def __init__(self):
         dai.Buffer.__init__(self)
         self._classes: List[str] = []
         self._scores: NDArray[np.float32] = np.array([])
-        self._transformation: dai.ImgTransformation = None
+        self._transformation: dai.ImgTransformation | None = None
```
klemen1999, line 30: Should be `Optional[...]` because `|` type annotation is not supported by py3.8

PR #284, opened 2026-02-19
Remove need for cls_scores = scores.max(axis=-1) as ReduceMax resides on the model side for YOLO26

dtronmans:
## Purpose
Also change expected names for yolo26 fields in the NNArchive

Depends on pruned models from https://github.com/luxonis/tools/pull/206


klemen1999 (approved):
LGTM but let's merge after tools merge

PR #299, opened 2026-06-17
Add RF-DETR instance segmentation parsing

rolandocortez:
## Purpose

Add RF-DETR instance segmentation support to `RFDETRParser`.

RF-DETR Seg models expose an additional `masks` output next to the existing `dets` and `labels` outputs. Before this change, the parser only produced detection boxes/classes and ignored segmentation masks, so RF-DETR Seg archives could run but would not emit an instance segmentation mask through `create_detection_message`.

## Specification

This PR extends `RFDETRParser` to support both RF-DETR detection and RF-DETR instance segmentation outputs.

Main changes:

* Supports either 2 outputs (`dets`, `labels`) or 3 outputs (`dets`, `labels`, `masks`).
* Adds `mask_conf` parser configuration for binarizing instance masks.
* Reads the model input shape from the NNArchive config and uses it as the final segmentation mask shape.
* Adds RF-DETR Seg mask post-processing:

  * sigmoid
  * crop mask to detection bbox
  * threshold using `mask_conf`
  * resize with `cv2.INTER_NEAREST`
  * compose final `uint8` instance mask with `255` as background
* Passes the composed mask through `create_detection_message(..., masks=final_mask)`.
* Keeps the existing detection-only path backward compatible.

This change adds an OpenCV import to `RFDETRParser` for resizing instance masks with `cv2.INTER_NEAREST`.

Potential impact is limited to RF-DETR parsing. Detection-only RF-DETR models should continue to work as before because segmentation logic is only used when the `masks` output is present.

## Deployment Plan

This should be released as part of the next `depthai-nodes` release. Rollback would be reverting this parser change.

## Testing & Validation

Local checks:

* `python -m py_compile depthai_nodes/node/parsers/rf_detr.py`
* `git diff --check`
* `python -m ruff check depthai_nodes/node/parsers/rf_detr.py`
* `pytest tests/unittests/test_creators/test_detections.py -q`

  * `27 passed`
* `pytest tests/unittests/test_creators/test_segmentation.py -q`

  * `5 passed`
* `pytest tests/unittests/test_creators -q`

  * `125 passed`

Additional smoke tests:

* RF-DETR Seg mask post-processing smoke test.
* RF-DETRParser build/config smoke test for 3-output segmentation config.

RVC4 validation:

* Validated RF-DETR Seg Nano archive on NUC with an RVC4 device and `ParsingNeuralNetwork`.
* Confirmed detections and `cvSegmentationMask` are present in output messages.
* Compared RVC4 outputs against PyTorch `RFDETRSegNano` on the same RVC4 passthrough clean frames.
* Image-level validation over 10 COCO source images / 40 sampled frames:

  * matched detections: `93.7%`
  * mean box IoU: `0.967`
  * mean mask IoU: `0.795`
  * masks present in all sampled frames

## AI Usage

Assisted-by: ChatGPT

## Summary by CodeRabbit

* **New Features**
  * RF-DETR parser now supports instance segmentation outputs, including per-instance mask generation.
  * Added a configurable mask confidence threshold (`mask_conf`) and a `setMaskConfidence()` control for fine-tuning segmentation results.

* **Improvements**
  * Masks are resized to match the model’s expected input resolution for better alignment.
  * Enhanced numerical stability in sigmoid/mask probability computation.
  * Improved segmentation post-processing and confidence filtering for consistent box/label/mask selection.

### DIFF
--- a/depthai_nodes/node/parsers/rf_detr.py
+++ b/depthai_nodes/node/parsers/rf_detr.py
@@ -1,13 +1,15 @@
 from typing import Any
 
-import cv2
 import depthai as dai
 import numpy as np
 
 from depthai_nodes.message.creators import create_detection_message
 from depthai_nodes.node.parsers.base_parser import BaseParser
-from depthai_nodes.node.parsers.utils import xyxy_to_xywh
-from depthai_nodes.node.parsers.utils.masks_utils import crop_mask
+from depthai_nodes.node.parsers.utils import sigmoid, xyxy_to_xywh
+from depthai_nodes.node.parsers.utils.bbox_format_converters import xywh_to_xyxy
+from depthai_nodes.node.parsers.utils.masks_utils import (
+    process_single_mask_rfdetr,
+)
 
 
 class RFDETRParser(BaseParser):
@@ -71,7 +73,7 @@ class RFDETRParser(BaseParser):
         self.output_layer_names: list[str] = []
         self.input_shape: tuple[int, int] | None = None
         self._logger.debug(
-            f"RFDETRParser initialized with conf_threshold={conf_threshold}, max_det={max_det}"
+            f"RFDETRParser initialized with conf_threshold={conf_threshold}, max_det={max_det}, mask_conf={mask_conf}"
         )
 
     @property
@@ -122,10 +124,10 @@ class RFDETRParser(BaseParser):
         self._logger.debug(f"Label names updated to: {label_names}")
 
     def setMaskConfidence(self, mask_conf: float) -> None:
-        """Set mask confidence threshold.
+        """Sets the mask confidence threshold.
 
-        RF-DETR Seg outputs mask logits. The parser applies sigmoid to the mask logits,
-        so this value is interpreted as a normal probability threshold.
+        @param mask_conf: The mask confidence threshold.
+        @type mask_conf: float
         """
         if not isinstance(mask_conf, float):
             raise ValueError("Mask confidence threshold must be a float.")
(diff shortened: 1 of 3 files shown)


depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -211,23 +308,44 @@ def run(self):
...
+
+                final_mask = np.full(self.input_shape, 255, dtype=np.uint8)
+
+                for i, (mask_logits, bbox) in enumerate(zip(masks, boxes_cxcywh)):
+                    resized_mask = self._process_mask(
+                        mask_logits,
+                        bbox,
+                        self.input_shape,
+                    )
+                    final_mask[resized_mask > 0] = i
```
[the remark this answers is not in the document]
klemen1999: Let's clamp max detections to 255 (also inside the tutorial, in the "making NNArchive" step where we set parser metadata). And if len(detections) is bigger than 255 let's issue a warning that some instances were ignored @rolandocortez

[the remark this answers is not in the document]
rolandocortez: Updated this in the latest commit

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -52,12 +60,16 @@ def __init__(
...
+        @type mask_conf: float
         """
         super().__init__()
         self.conf_threshold = conf_threshold
         self.max_det = max_det
         self.label_names = label_names
+        self.mask_conf = mask_conf
         self.output_layer_names: list[str] = []
+        self.input_shape: tuple[int, int] | None = None
         self._logger.debug(
```
klemen1999, line 75: Let's mask_conf in the log

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -109,6 +121,21 @@ def setLabelNames(self, label_names: list[str]) -> None:
         self.label_names = label_names
         self._logger.debug(f"Label names updated to: {label_names}")
 
+    def setMaskConfidence(self, mask_conf: float) -> None:
+        """Set mask confidence threshold.
+
+        RF-DETR Seg outputs mask logits. The parser applies sigmoid to the mask logits,
```
klemen1999: I don't think we ned to be this explicit here. I would literally copy paste the costring from YOLOExtendedParser setMaskConfidence() function

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -149,7 +199,16 @@ def _sigmoid(x: np.ndarray) -> np.ndarray:
         @return: Sigmoid of input.
         @rtype: np.ndarray
         """
-        return 1 / (1 + np.exp(-x))
+        x = np.asarray(x, dtype=np.float32)
```
klemen1999: So as I understand this is a numerically more stable implementation of sigmoid comapred to [this sigmoid implementation](https://github.com/luxonis/depthai-nodes/blob/e067e90a63d8a1b1e05715b39e862e80a30d39a4/depthai_nodes/node/parsers/utils/activations.py#L27-L35).
If that is the case and these two are actually practically interchangeable then I would prefer we change the sigmoid implementation that we already have to this one and here just reuse that function (import and use)

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -149,7 +199,16 @@ def _sigmoid(x: np.ndarray) -> np.ndarray:
...
+        positive_mask = x >= 0
+        result[positive_mask] = 1.0 / (1.0 + np.exp(-x[positive_mask]))
+
+        exp_x = np.exp(x[~positive_mask])
+        result[~positive_mask] = exp_x / (1.0 + exp_x)
+
+        return result
 
     @staticmethod
     def _box_cxcywh_to_xyxy(boxes: np.ndarray) -> np.ndarray:
```
klemen1999: I think we should be able to use already implemented [xywh_to_xyxy()](https://github.com/luxonis/depthai-nodes/blob/e067e90a63d8a1b1e05715b39e862e80a30d39a4/depthai_nodes/node/parsers/utils/bbox_format_converters.py#L115-L116) function here instead of re-implementing it?

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -169,6 +228,42 @@ def _box_cxcywh_to_xyxy(boxes: np.ndarray) -> np.ndarray:
         ymax = cy + h / 2
         return np.stack([xmin, ymin, xmax, ymax], axis=-1)
 
+    def _process_mask(
```
klemen1999: I'd put this function in [masks_utils](https://github.com/luxonis/depthai-nodes/blob/e067e90a63d8a1b1e05715b39e862e80a30d39a4/depthai_nodes/node/parsers/utils/masks_utils.py#L1-L2) under name e.g. `process_single_mask_rfdetr()` and import it from there

depthai_nodes/node/parsers/rf_detr.py
```diff
@@ -203,31 +227,75 @@ def run(self):
...
+                final_mask = np.full(self.input_shape, 255, dtype=np.uint8)
+
+                for i, (mask_logits, bbox) in enumerate(zip(masks, boxes_cxcywh)):
+                    resized_mask = process_single_mask_rfdetr(
+                        mask_logits=mask_logits,
+                        mask_conf=self.mask_conf,
+                        bbox=bbox,
+                        input_shape=self.input_shape,
+                    )
+                    final_mask[resized_mask > 0] = i
```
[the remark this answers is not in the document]
rolandocortez: Fixed. I now only fill background pixels so lower-score masks don’t overwrite earlier ones.

rolandocortez:
Pushed the cleanup updates.

I moved the RF-DETR mask post-processing into `masks_utils`, reused the shared bbox conversion helper, and switched RF-DETR to the shared sigmoid implementation after making it numerically stable. I also addressed the mask overlap case so lower-score masks don’t overwrite higher-score ones.

I kept the new helper RF-DETR-specific for now since it’s the only current caller, but we can generalize it later if another parser needs the same direct-mask-logits flow.

klemen1999 (approved):

PR #301, opened 2026-06-26
Use YOLO metadata strides in parser

rolandocortez:
## Purpose

Support YOLO-like models with non-default head stride layouts, including `n_heads = 4`.

Previously, `YOLOExtendedParser` used hard-coded fallback strides depending on the YOLO subtype. This meant archives with 4 output heads could be parsed with only the legacy 3-stride fallback, silently ignoring the extra stride scale.

This change lets the parser consume stride metadata exported in NNArchives, while preserving the existing fallback behavior for older archives.

## Specification

* Added metadata-driven stride resolution for `YOLOExtendedParser`.
* The parser now reads `strides` from the NNArchive head metadata when present.
* Legacy fallback behavior is preserved when `strides` is not present:

  * default YOLO fallback: `[8, 16, 32]`
  * tiny YOLO fallback: `[16, 32]`
* Explicit metadata strides are validated:

  * must be a list or tuple
  * must not be empty
  * values must be positive integers
  * number of strides must match the number of YOLO outputs

Paired with the `luxonis-train` PR that exports `metadata.strides` for YOLO-like detection heads.

Backward compatibility is preserved for existing NNArchives without stride metadata. Archives that provide invalid explicit stride metadata will now fail fast with a clear error instead of being parsed with an incorrect stride configuration.

## Deployment Plan

Rollback is straightforward: revert this parser change and the parser will return to the previous hard-coded fallback behavior.

## Testing & Validation

Validated with task-specific NNArchives generated from `luxonis-train` and converted to RVC2 via HubAI.

Cases validated:

* 4 heads: `metadata.strides = [4, 8, 16, 32]`
* 2 heads: `metadata.strides = [16, 32]`
* 3 heads with offset `attach_index`: `metadata.strides = [4, 8, 16]`

Existing `depthai-nodes` E2E validation:

* Ran `tests/end_to_end/test_e2e.py::test_pipelines`
* Used the three generated RVC2 NNArchives above
* Ran on a real RVC2 device
* Result: `3 passed in 44.71s`

Also validated existing YOLO parser coverage:

* YOLO parser/generator/parsing tests passed: `3 passed`

## AI Usage

Assisted-by: ChatGPT

## Summary by CodeRabbit

* **New Features**
  * Added configurable YOLO head strides via model configuration, with shared stride resolution during decoding.
  * When strides differ from expected defaults, YOLO heads are routed to the appropriate parsing path for correct results.

* **Bug Fixes**
  * Added stricter stride validation (type, non-empty, positive integers, and expected count match).
  * Improved debug output to include the resolved strides for easier troubleshooting.

### DIFF
--- a/depthai_nodes/node/parsers/utils/yolo.py
+++ b/depthai_nodes/node/parsers/utils/yolo.py
@@ -34,6 +34,52 @@ class YOLOSubtype(str, Enum):
     DEFAULT = ""
 
 
+def resolve_yolo_strides(
+    strides: list[int] | tuple[int, ...] | None,
+    subtype: YOLOSubtype,
+    num_outputs: int,
+) -> list[int]:
+    """Resolve YOLO strides from metadata or default fallback.
+
+    @param strides: Optional strides from NNArchive head metadata.
+    @type strides: list[int] | tuple[int, ...] | None
+    @param subtype: YOLO subtype.
+    @type subtype: YOLOSubtype
+    @param num_outputs: Number of YOLO output heads.
+    @type num_outputs: int
+    @return: Resolved YOLO strides.
+    @rtype: list[int]
+    """
+    if strides is None:
+        return (
+            [8, 16, 32]
+            if subtype not in [YOLOSubtype.V3UT, YOLOSubtype.V3T, YOLOSubtype.V4T]
+            else [16, 32]
+        )
+
+    if not isinstance(strides, (list, tuple)):
+        raise ValueError("`strides` must be a list or tuple of positive integers.")
+
+    if not strides:
+        raise ValueError("`strides` must not be empty.")
+
+    resolved_strides = list(strides)
+
+    if not all(type(stride) is int for stride in resolved_strides):
+        raise ValueError("All `strides` values must be integers.")
+
+    if not all(stride > 0 for stride in resolved_strides):
+        raise ValueError("All `strides` values must be positive.")
+
+    if len(resolved_strides) != num_outputs:
+        raise ValueError(
+            "Number of `strides` must match number of YOLO outputs. "
+            f"Got {len(resolved_strides)} strides for {num_outputs} outputs."
+        )
+
+    return resolved_strides
+
+
 def make_grid_numpy(ny: int, nx: int, na: int) -> np.ndarray:
     """Create a grid of shape (1, na, ny, nx, 2)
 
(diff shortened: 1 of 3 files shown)


depthai_nodes/node/parsers/utils/yolo.py
```diff
@@ -34,6 +34,55 @@ class YOLOSubtype(str, Enum):
...
+    @param strides: Optional strides from NNArchive head metadata.
+    @type strides: list[int] | tuple[int, ...] | None
+    @param subtype: YOLO subtype.
+    @type subtype: YOLOSubtype
+    @param num_outputs: Number of YOLO output heads.
+    @type num_outputs: int
+    @return: Resolved YOLO strides.
+    @rtype: list[int]
+    """
+    if strides is None:
```
dtronmans, line 53: One problem we can encounter now is that there is routing logic in the [ParserGenerator](https://github.com/luxonis/depthai-nodes/blob/d98ce6d9f04862c32670472a012912c474984e5c/depthai_nodes/node/parser_generator.py#L75-L96) that defaults to a native `dai.node.DetectionParser` if the `subtype` field in the `config.json` is recognized as a [dai-supported subtype](https://github.com/luxonis/depthai-nodes/blob/d98ce6d9f04862c32670472a012912c474984e5c/depthai_nodes/node/parser_generator.py#L18), even if the `parser` field is explicitly set to `YOLOExtendedParser`. This is logic that I was not aware of until now when I tried to run a 4-stride NNArchive using the [generic-example](https://github.com/luxonis/oak-examples/tree/main/neural-networks/generic-example), because `YOLO` and `YOLOExtendedParser` behaviors are basically the same.

The error I get is a DAI-native error that signals a mismatch between strides and outputs. Because we need to also make the change on the DAI-side, I think for now we need to find some way to temporarily route to `YOLOExtendedParser` in the `ParserGenerator` if all 3 are true:

- YOLOExtendedParser is set in the config
- The `subtype` is a DAI-supported subtype
- `strides` is set (or is set to a non-default, like [4, 8, 16 ,32]

Let me know if you have any questions about this because I was a bit confused by the distinction initially!

rolandocortez: Thanks Dani, great catch. Didn't notice that ParserGenerator routing detail initially.

I pushed an update for this. `YOLOExtendedParser` configs with DAI-supported subtypes and non-default explicit `strides` now stay on the host-side `YOLOExtendedParser` path instead of being routed to `dai.node.DetectionParser`.

I also applied the same guard to `parser="YOLO"` archives with non-default explicit `strides`, since those hit the `DEVICE_PARSERS` path directly and would otherwise bypass the host-side parser as well. Default/no-stride YOLO layouts still keep the existing native `DetectionParser` route.

I validated the routing with a before/after check:
- Before: `YOLO` and `YOLOExtendedParser` configs with non-default strides were routed to `DetectionParser`.
- After: non-default strides route to `YOLOExtendedParser`, while default explicit `[8, 16, 32]` still routes to `DetectionParser`.

I also ran Klemen’s compatibility check separately: YOLOv6 Nano from Zoo with manually added `metadata.strides = [8, 16, 32]` runs with explicit `dai.node.DetectionNetwork` without crashing.

dtronmans: Thanks Rolando, I just tested it out again and it works correctly!

dtronmans (approved):

klemen1999 (approved):

PR #303, opened 2026-07-01
Add sync subnode for FrameCropper inputs

tadeas0:
## Purpose
Add Sync subnode in front of `FrameCropper` inputs. This enables non-blocking queue usage earlier in the pipeline and makes the `FrameCropper` node more robust.

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation
All oak-examples using `FrameCropper` work same as before after this change.

## AI Usage

Assisted-by: Pi Coding Agent:GPT-5.5

## Summary by CodeRabbit

* **New Features**
  * Added an optional synchronization threshold (`syncThreshold`) for frame cropping to better pair frames with detections.
  * Enhanced waiting-on-config workflows for frame processing with manipulation configs, including optional synchronization when `waitForConfig` is enabled.

* **Bug Fixes**
  * Improved pairing reliability by synchronizing frame and related metadata into a single matched stream, reducing mismatches between images and their associated detections/configs.


depthai_nodes/node/frame_cropper.py
```diff
@@ -178,12 +198,15 @@ def fromManipConfigs(
         inputManipConfigs: dai.Node.Output,
         maxOutputFrameSize: int,
         waitForConfig: bool,
+        syncThreshold: timedelta = timedelta(milliseconds=10),
```
PetrNovota: I have a small issue with this. The situation here is that if `waitForConfig == False` then the `syncThreshold` doesn't matter, is not used.
I always try to avoid a situation like this where two or more arguments are dependent on one another in this way.

I can notice that the `syncThreshold` arg is also needed in the other fluent method, `fromImgDetections`.

One way of dealing with this could be  to pull this arg out into its own fluent method:
`def withSyncThreshold(self, syncThreshold: timedelta)`

but might be an overkill. In any case, I would introduce a check when `waitForConfig` is set to `False` and the `syncThreshold` is set by the user and warn him that the syncThreshold will not be used.

depthai_nodes/node/frame_cropper.py
```diff
@@ -225,7 +249,12 @@ def run(self) -> None:
         return  # Both fromImgDetections and fromManipConfigs use on-device Script
 
     def _build_detections_cropper(self, input_image: dai.Node.Output):
-        self._script = self._pipeline.create(dai.node.Script)
+        assert self._input_img_detections is not None
```
PetrNovota: I think there are a couple of issues with these assertions:
1. They are only used internally by the `build` method and the method already checks if one of the fluent builder methods was used
```
        if not self._version_selected:
            raise RuntimeError(
                "Configure the FrameCropper by calling one of the `fromImgDetections` or `fromManipConfigs` methods first."
            )
```
This was the place where I wanted to check whether the build method is not called in wrong order. And these asserts serve the same purpose. 
2. If we keep the asserts, I think we must add an explanation text, as it stands right now, user would get an unintelligible error, it would be unclear what is wrong.

PetrNovota (commented):
Thank you. Looking good. Left a couple of minor comments

depthai_nodes/node/frame_cropper.py
tadeas0: Added assert messages

depthai_nodes/node/frame_cropper.py
tadeas0: Added check for `syncThreshold` and `waitForConfig == False` exclusivity.

PetrNovota (approved):
lgtm

PR #304, opened 2026-07-03
Use native segmentation parser and mask

rolandocortez:
## Purpose
Use native DepthAI segmentation components in `depthai-nodes`.

This PR updates the segmentation flow to use native `dai.SegmentationMask` and native `dai.node.SegmentationParser` where applicable, while preserving the existing host-side `depthai-nodes` parser for `hostOnly=True`.

It also aligns host parser background semantics with the native parser and fixes visualization of binary segmentation masks where class `0` is a valid foreground class and `255` is background.

Additionally, this PR adds the missing `setStrides(...)` setter to `YOLOExtendedParser`.

## Specification
- Removed the custom `depthai_nodes.message.segmentation.SegmentationMask` message.
- Added explicit `copy_message(dai.SegmentationMask)` support.
- Updated `ParserGenerator` to use native `dai.node.SegmentationParser` when `parser_name == "SegmentationParser"` and `hostOnly=False`.
- Preserved `depthai_nodes.node.parsers.segmentation.SegmentationParser` for `hostOnly=True`.
- Added RVC2 handling for native segmentation parser by setting it to run on host.
- Added `background_class` support to the host `SegmentationParser`.
- Updated host parser semantics to match native behavior:
  - `background_class=True` maps class `0` to `255`.
  - Foreground class ids are preserved.
  - Single-channel/unassigned-class behavior still shifts correctly.
  - `classes_in_one_layer=True` ignores `background_class`.
- Updated `ApplyColormap` handling for native `dai.SegmentationMask`:
  - `255` background maps to visual background index `0`.
  - class `0` maps to a visible foreground index.
- Added `YOLOExtendedParser.setStrides(...)` for manually setting YOLO output head strides.

This depends on a DepthAI version exposing native `dai.SegmentationMask` and `dai.node.SegmentationParser`.

Potential impact is limited to segmentation parsing, segmentation mask copying, segmentation visualization paths, and the new YOLO parser strides setter API. The existing host segmentation parser path remains available through `hostOnly=True`.

Direct Visualizer rendering of binary `dai.SegmentationMask` values `[0, 255]` may not visibly color class `0`; the `ApplyColormap -> ImgFrameOverlay` path handles this correctly.

## Deployment Plan

Rollback: revert this PR.

## Testing & Validation
- Focused unit tests passed:
  - `tests/unittests/test_messages/test_copy_message.py`
  - `tests/unittests/test_nodes/test_host_nodes/test_apply_colormap_node.py`
- Validated host `SegmentationParser.run()` directly with synthetic NN output.
- Confirmed host background semantics match native parser behavior:
  - class `0 -> 255` when `background_class=True`
  - foreground class ids are preserved
  - single-channel/unassigned-class path still shifts correctly
  - `classes_in_one_layer=True` ignores `background_class`
- Validated with the segmentation MRE on OAK-1 / RVC2:
  - DeepLab 512 host output no longer has foreground off-by-one compared with native parser.
  - DeepLab person host/native outputs produce valid `[0, 255]` masks.
  - `ApplyColormap -> ImgFrameOverlay` correctly visualizes binary masks where foreground class id is `0`.
- Validated `YOLOExtendedParser.setStrides(...)` locally:
  - accepts list and tuple stride values
  - rejects invalid, empty, non-integer, zero, and negative stride values
- `python -m compileall` passed for touched parser files.

Note: For `luxonis/mediapipe-selfie-segmentation:256x144`, a dual-parser MRE feeding the same `NeuralNetwork` output to both parsers showed native `dai.node.SegmentationParser` emitting only `[0]`, while the host `depthai-nodes` parser emitted `[0, 255]` for the same sequence numbers and mask shape. Since both parsers consumed the same NN output, this appears isolated to the native parser/model metadata path rather than the host parser changes in this PR.

## AI Usage
Assisted-by: ChatGPT

## Summary by CodeRabbit

* **New Features**
  * Added native DepthAI `dai.SegmentationMask` support across parsing/remapping and ApplyColormap processing.
  * Added `background_class` option to segmentation outputs.
  * Added `setStrides` support for extended YOLO parsing.
  * Updated message documentation to introduce `SnapData` (replacing the previous segmentation-mask documentation).
* **Bug Fixes**
  * Standardized segmentation masks to `uint8`, with consistent handling for empty and unassigned pixels.
* **Documentation**
  * Updated parser/node docs and type references to `dai.SegmentationMask`.
* **Tests**
  * Updated unit, stability, and relevant node tests to validate the native mask API (including mask copying behavior).
* **Chores**
  * Bumped DepthAI version to `3.8.0` in workflows and dependency constraints.


depthai_nodes/message/utils/copy_message.py
```diff
@@ -49,6 +51,18 @@ def _copy_img_frame(img_frame: dai.ImgFrame) -> dai.ImgFrame:
...
 
+    def _copy_segmentation_mask(
+        segmentation_mask: dai.SegmentationMask,
+    ) -> dai.SegmentationMask:
+        segmentation_mask_copy = _copy_metadata(segmentation_mask)
+        assert isinstance(segmentation_mask_copy, dai.SegmentationMask)
+        segmentation_mask_copy.setCvMask(segmentation_mask.getCvMask().copy())
+        if hasattr(segmentation_mask, "getLabels") and hasattr(
+            segmentation_mask_copy, "setLabels"
+        ):
```
klemen1999: Why do we have this defensive programming here? The `dai.SegmentationMask` has defined API so either there is `getLabels` and `setLabels` or there isn't. Given that this deptahi-nodes version will be compatible only with DAI>=3.7.1 we don't have to be so defensive with "hasattr"

depthai_nodes/node/parsers/utils/fastsam.py
```diff
@@ -386,7 +386,7 @@ def merge_masks(masks: np.ndarray) -> np.ndarray:
     else:
         raise ValueError("Masks must be a 3D array.")
 
-    merged_masks = np.full((height, width), -1, dtype=np.int16)
+    merged_masks = np.full((height, width), 255, dtype=np.uint8)
```
klemen1999, line 389: We should correct the FastSam test so it expects uint8 instead of uin16 (related to currently failing integration test)

depthai_nodes/node/parsers/fastsam.py
```diff
@@ -378,7 +378,7 @@ def run(self):
                     )
 
             if len(results_masks) == 0:
-                results_masks = np.full((1, height, width), -1, dtype=np.int16)
+                results_masks = np.full((1, height, width), 255, dtype=np.uint8)
             results_masks = merge_masks(results_masks)
```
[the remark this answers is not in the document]
klemen1999: Index 255 is semantically the index of background/unassigned pixel

depthai_nodes/node/parsers/segmentation.py
```diff
@@ -92,9 +112,16 @@ def build(
         self.classes_in_one_layer = head_config.get(
             "classes_in_one_layer", self.classes_in_one_layer
         )
+        self.background_class = head_config.get(
+            "background_class",
+            head_config.get("backgroundClass", self.background_class),
```
klemen1999: The nn_archive param is `background_class`, `backgroundClass` is not valid and we shouldn't look for it.

klemen1999 (commented):
Few comments but overall looks good

depthai_nodes/node/parsers/segmentation.py
```diff
@@ -41,12 +44,18 @@ def __init__(
...
         super().__init__()
         self.output_layer_name = output_layer_name
         self.classes_in_one_layer = classes_in_one_layer
+        self.background_class = background_class
         self._logger.debug(
-            f"SegmentationParser initialized with output_layer_name='{output_layer_name}', classes_in_one_layer={classes_in_one_layer}"
+            "SegmentationParser initialized with "
+            f"output_layer_name='{output_layer_name}', "
+            f"classes_in_one_layer={classes_in_one_layer}, "
+            f"background_class={background_class}"
```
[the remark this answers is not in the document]
klemen1999: @rolandocortez let's check how DAI native SegmentationParser handles case when `background_class=True and classes_in_one_layer=True`: Does it crash or warn or does nothing and just ignore?
In our case I think it would be best to warn once and say that background_class=True will be ignored because classes_in_one_layer=True. This way we inform the user about the behaviour but not raise an error to crash the pipeline

klemen1999:
TODO: Before merge we need next DepthAI version released. This means that we need to wait with merge and update the requirements.txt

depthai_nodes/node/parser_generator.py
```diff
@@ -100,6 +100,18 @@ def build(
                         parsers[index] = parser
                         continue
 
+            elif parser_name == "SegmentationParser" and not hostOnly:
+                parser = pipeline.create(dai.node.SegmentationParser)
+                parser.setNNArchive(nnArchive)
```
klemen1999: Actually here we can switch to using `.setNNArchiveHead()` which would take `head` as input.
And let's do the same in the code above to use `.setNNArchiveHead()` for `DetectionParser` usage as well

rolandocortez:
Got it. Pushed the follow-up fixes and CI is green now. I fixed the stability failures from the native SegmentationMask migration, addressed the review comments around getLabels/setLabels, background_class handling, and setNNArchiveHead(head), and updated the parser mock to match the new ParserGenerator API usage.

depthai_nodes/node/parsers/fastsam.py
```diff
@@ -378,7 +378,7 @@ def run(self):
                     )
 
             if len(results_masks) == 0:
-                results_masks = np.full((1, height, width), -1, dtype=np.int16)
+                results_masks = np.zeros((1, height, width), dtype=np.uint8)
```
klemen1999: Shouldn't here be np.full with 255 values? Since before it was -1 and -1 before meant "all background" while now 255 means "all background" (0 would mean actually 0th class)

tests/stability_tests/check_messages.py
```diff
@@ -106,8 +105,16 @@ def check_embeddings_msg(
     np.testing.assert_allclose(embeddings, expected_embeddings, rtol=1e-2)
 
 
+def _normalize_expected_segmentation_mask(expected_mask: np.ndarray) -> np.ndarray:
+    if expected_mask.dtype == np.int16:
+        expected_mask = np.where(expected_mask == -1, 255, expected_mask).astype(
+            np.uint8
+        )
+    return expected_mask
```
klemen1999: This shouln't be needed. Instead we should correct the expected outputs from [here](https://github.com/luxonis/depthai-nodes/blob/f7e3814928b195878e2f472543e615fd0de23fbf/tests/stability_tests/check_messages.py#L20-L22)

tests/stability_tests/run_parser_test.py
```diff
@@ -90,7 +90,10 @@ def test_parser(parser_generator, model: str, parser_name: str, duration: int):
         exit(9)
 
     # Create and build parser
-    parser = parser_generator.build(nnArchive=nn_archive)[0]
+    parser = parser_generator.build(
+        nnArchive=nn_archive,
+        hostOnly=parser_name == "SegmentationParser",
+    )[0]
```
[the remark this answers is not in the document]
klemen1999: dai.node.SegmentationParser is already tested in DepthAI library

tests/stability_tests/check_messages.py
klemen1999: I changed the "ground truth" values on the bucket. You should be now able to remove this function completely and the tests as they were before should work just fine

rolandocortez: Removed the normalization helper and pushed the FastSAM fallback update.

CI now passes FastSAM and MediaPipe segmentation, but DeepLab still fails because the downloaded expected mask is still int16 while the parser emits native uint8. Looks like the bucket expected output for luxonis/deeplab-v3-plus:256x256 may still be in the old format. Could you check/update that one too? I can rerun after that.

.github/workflows/e2e_tests.yaml
```diff
@@ -111,7 +111,8 @@ jobs:
...
             fi
           else
-            exec hil_runner --platforms $PLATFORM --wait --reservation-name $RESERVATION_NAME --before-docker-pull "echo ${{ github.token }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin" --docker-image ghcr.io/luxonis/depthai-nodes-testing --docker-run-args "--env LUXONIS_EXTRA_INDEX_URL=${{secrets.LUXONIS_EXTRA_INDEX_URL}} --env DEPTHAI_VERSION=\"3.3.0\" --env HUBAI_TEAM_SLUG=${{ secrets.HUBAI_TEAM_SLUG }} --env HUBAI_API_KEY=${{ secrets.HUBAI_API_KEY }} --env BRANCH=${{ github.ref_name }} --env FLAGS=\"-all --platform ${{ matrix.platform }}\""
+            exec hil_runner --platforms $PLATFORM --wait --reservation-name $RESERVATION_NAME --before-docker-pull "echo ${{ github.token }} | docker login ghcr.io -u ${{ github.actor }} --password-stdin" --docker-image ghcr.io/luxonis/depthai-nodes-testing --docker-run-args "--env LUXONIS_EXTRA_INDEX_URL=${{secrets.LUXONIS_EXTRA_INDEX_URL}} --env DEPTHAI_VERSION=\"3.8.0\" --env HUBAI_TEAM_SLUG=${{ secrets.HUBAI_TEAM_SLUG }} --env HUBAI_API_KEY=${{ secrets.HUBAI_API_KEY }} --env BRANCH=${{ github.ref_name }} --env FLAGS=\"-all --platform ${{ matrix.platform }}\""
+
```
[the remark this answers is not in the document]
rolandocortez: Addressed in d89ad06.

Moved `github.ref_name` into step-level env variables and now reference quoted shell variables in the E2E push/default paths instead of interpolating `github.ref_name` directly inside the shell command string.

Also applied the same pattern to the Windows E2E path.

klemen1999 (approved):

PR #319, opened 2026-08-21
Isolate macOS end-to-end test environment

tjb7:
This change makes the macOS end-to-end test environment deterministic and isolated from the testbed’s pre-existing .hil virtual environment.

  - Uses uv 0.12.5 to create a managed Python 3.12 project virtual environment.
  - Installs and runs depthai-nodes and test dependencies through that environment explicitly.
  - Launches child test processes with sys.executable, avoiding bare python resolution to the .hil interpreter.
  - Pins the outer HIL bootstrap pip version to 26.2.1 and invokes it as python -m pip.

  This prevents the project tests from silently using the testbed’s shared HIL Python environment.

## Summary by CodeRabbit

* **Tests**
  * Improved end-to-end test reliability with managed Python 3.12 environments.
  * Standardized test execution using the required `uv` version.
  * Improved subprocess handling and interpreter consistency during pipeline tests.
  * Added clearer diagnostics for test environments and command execution.

* **Chores**
  * Updated installation steps for consistent package management.
  * Improved Linux, Windows, and macOS test setup and command availability.
  * Enhanced reservation and environment configuration across supported platforms.

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -84,7 +84,7 @@ jobs:
 
       - name: Install HIL
         run: |
-          python -m pip install --upgrade pip==26.2.1
+          python -m pip install --upgrade pip
           python -m pip --version
            python -m pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple
       - name: Run Test
@@ -139,7 +139,7 @@ jobs:
 
       - name: Install HIL
         run: |
-          python -m pip install --upgrade pip==26.2.1
+          python -m pip install --upgrade pip
           python -m pip --version
            python -m pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple
       - name: Run Test
@@ -195,7 +195,7 @@ jobs:
           cd depthai-nodes
           python3 -m venv venv
           . venv/bin/activate
-          python -m pip install --upgrade pip==26.2.1
+          python -m pip install --upgrade pip
           python -m pip --version
           python -m pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple --no-cache-dir
           JOB=${{ job.check_run_id }}
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -76,6 +76,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.12
+          cache: pip
+
+      - name: Install HIL
+        run: |
+          python -m pip install --upgrade pip==26.2.1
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e_tests.yaml
```diff
@@ -198,10 +204,10 @@ jobs:
                 ADDITIONAL_OPTIONS="${{ github.event.inputs.additional_options }}"
             fi
             if [[ "${{ github.event.inputs.bom_mode }}" == 'true' ]]; then
-              exec hil_runner --hold-reservation $ADDITIONAL_OPTIONS -os mac --basic-sanity --models 'oak4_s or oak4_pro or oak4_d' --wait $RESERVATION_OPTION --sync-workspace --commands "cd /tmp/depthai-nodes && ./tests/end_to_end/run_end_to_end.sh ${{ secrets.HUBAI_API_KEY }} ${{ secrets.HUBAI_TEAM_SLUG }} ${{ github.event.inputs.depthai-version }} ${{secrets.LUXONIS_EXTRA_INDEX_URL}} rvc4"
+              exec hil_runner --hold-reservation $ADDITIONAL_OPTIONS -os mac --basic-sanity --models 'oak4_s or oak4_pro or oak4_d' --wait $RESERVATION_OPTION --sync-workspace --commands "export PATH=\"\$HOME/.local/bin:\$PATH\"; cd /tmp/depthai-nodes && ./tests/end_to_end/run_end_to_end.sh ${{ secrets.HUBAI_API_KEY }} ${{ secrets.HUBAI_TEAM_SLUG }} ${{ github.event.inputs.depthai-version }} ${{secrets.LUXONIS_EXTRA_INDEX_URL}} rvc4"
```
tjb7, line 214: this change makes uv discoverable

tjb7:
Here is a run: https://github.com/luxonis/depthai-nodes/actions/runs/32478785714/job/96760533080

.github/workflows/e2e_tests.yaml
```diff
@@ -84,7 +84,9 @@ jobs:
 
       - name: Install HIL
         run: |
-          pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple
+          python -m pip install --upgrade pip==26.2.1
```
tjb7: This isn't great, as the environments should be upgraded separately from individual test runs ... the runs should just check that the proper version is in place. Perhaps there is a better way ....

danilo-pejovic: Previous step sets up python for this run - thus python needs to be set up in this step. I would not hardcode pip version and just upgrade it to latest.

danilo-pejovic (commented):
Please split this pull request into 2 pull requests - one with the fix for pip install and one for mac tests. 

https://github.com/luxonis/depthai-nodes/actions/runs/32478785714

https://github.com/luxonis/depthai-nodes/actions/runs/32473163847

Considering these 2 tests both run on same commit and have wildly different results and IP/MXID discovery still fails I would say we are nowhere near being certain that mac issue is done away with. No need to delay pip install fix due to that.

.github/workflows/e2e_tests.yaml
tjb7: changes made.

tjb7:
Latest pipeline is running here with both changes, seems fine, changes will be 'split out' into 2 PR's per reviewer request: https://github.com/luxonis/depthai-nodes/actions/runs/32934457076/job/98072908921

PR #322, opened 2026-08-26
tests: isolate macOS e2e environment

tjb7:
Depends on PR 321

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -197,7 +197,9 @@ jobs:
           cd depthai-nodes
           python3 -m venv venv
           . venv/bin/activate
-          pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple --no-cache-dir
+          python -m pip install --upgrade pip
+          python -m pip --version
+          python -m pip install hil_framework --upgrade --index-url https://__token__:${{ secrets.GITLAB_TOKEN }}@gitlab.luxonis.com/api/v4/projects/213/packages/pypi/simple --no-cache-dir
           JOB=${{ job.check_run_id }}
           export RESERVATION_NAME="https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}/job/${JOB}"
           if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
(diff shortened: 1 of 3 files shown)


tests/end_to_end/run_end_to_end.sh
```diff
@@ -49,16 +49,33 @@ export FLAGS
 export DEPTHAI_NODES_LEVEL="debug"
 export DEPTHAI_DEBUG="0"
 
-python3.12 -m venv venv
+readonly UV_REQUIRED_VERSION="0.12.5"
```
danilo-pejovic, line 52: This script is used for both linux and mac testing - I assume this will break linux testing as we dont want UV there. We should likely separate it into 2 scripts or make sure script works for both.

.github/workflows/e2e_tests.yaml
```diff
@@ -211,10 +213,10 @@ jobs:
                 ADDITIONAL_OPTIONS="${{ github.event.inputs.additional_options }}"
             fi
             if [[ "${{ github.event.inputs.bom_mode }}" == 'true' ]]; then
-              exec hil_runner --hold-reservation $ADDITIONAL_OPTIONS -os mac --basic-sanity --models 'oak4_s or oak4_pro or oak4_d' --wait $RESERVATION_OPTION --sync-workspace --commands "cd /tmp/depthai-nodes && ./tests/end_to_end/run_end_to_end.sh ${{ secrets.HUBAI_API_KEY }} ${{ secrets.HUBAI_TEAM_SLUG }} ${{ github.event.inputs.depthai-version }} ${{secrets.LUXONIS_EXTRA_INDEX_URL}} rvc4"
+              exec hil_runner --hold-reservation $ADDITIONAL_OPTIONS -os mac --basic-sanity --models 'oak4_s or oak4_pro or oak4_d' --wait $RESERVATION_OPTION --sync-workspace --commands "export PATH=\"\$HOME/.local/bin:\$PATH\"; cd /tmp/depthai-nodes && ./tests/end_to_end/run_end_to_end.sh ${{ secrets.HUBAI_API_KEY }} ${{ secrets.HUBAI_TEAM_SLUG }} ${{ github.event.inputs.depthai-version }} ${{secrets.LUXONIS_EXTRA_INDEX_URL}} rvc4"
```
danilo-pejovic, line 216: Will this be removed when UV is properly installed to testbeds?

tests/end_to_end/test_e2e.py
```diff
@@ -71,20 +72,21 @@ def test_pipelines(IP: str, ip_platform: str, nn_archive_path, model):
             f"Testing model {model} from NN archive {nn_archive_path} on device with IP {IP} ({ip_platform})",
             flush=True,
         )
+        command = [sys.executable, "manual.py"]
```
danilo-pejovic, line 75: I guess this is the same as what it was before? Any reason for these changes?

danilo-pejovic (commented):
Lets run it on all OS-es, just so we make sure we are not breaking anything

PR #329, opened 2026-09-03
Change order of mask post-processing: threshold after interpolating instead of the reverse

dtronmans:
## Purpose
- Follow Mask RCNN, Ultralytics, Mask2Former and other popular libraries' steps to post-process masks by thresholding after upsampling instead of the reverse, which is what was previously being done.
- Should lead to smoother masks at the boundaries and less loss of information from the threshold that happens before resizing
- Tested using LuxonisEval regression test that asserts that ONNX and RVC4 FP16 NNArchive return very close instance segmentation metrics to LuxonisTrain, which is currently (correctly) thresholding after upsampling

## Specification

None / not applicable

None / not applicable

## Deployment Plan

None / not applicable

## Testing & Validation

None / not applicable

## AI Usage

Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]

Submitted code was reviewed by a human: YES/NO

The author is taking the responsibility for the contribution: YES/NO

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved segmentation mask generation across supported detection models.
  * Masks now align more reliably with requested output dimensions and bounding boxes.
  * Corrected confidence threshold handling for more accurate mask boundaries.
  * Improved mask processing consistency across YOLO and FastSAM workflows.
  * Preserved coordinate normalization and keypoint labels when copying message data.

### DIFF
--- a/depthai_nodes/message/utils/copy_message.py
+++ b/depthai_nodes/message/utils/copy_message.py
@@ -29,7 +29,7 @@ def copy_message(msg: dai.Buffer) -> dai.Buffer:
         raise TypeError(f"Copying of message type {type(msg)} is not supported.") from e
 
 
-def _copy(msg: dai.Buffer) -> dai.Buffer:
+def _copy(msg: dai.Buffer) -> dai.Buffer | object:
     def _copy_metadata(msg: dai.Buffer) -> dai.Buffer:
         msg_type = type(msg)
         msg_copy = msg_type()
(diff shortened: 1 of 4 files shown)


depthai_nodes/node/parsers/utils/masks_utils.py
```diff
@@ -34,6 +34,7 @@ def process_single_mask(
     mask_coeff: np.ndarray,
     mask_conf: float,
     bbox: np.ndarray,
+    output_shape: tuple[int, int],
```
klemen1999, line 44: Nipitck: Let's add in the docstring that this returns the mask in "output_shape" so it's clear that we perform resize as well

depthai_nodes/message/utils/copy_message.py
klemen1999, line 1: Why is this? Maybe we just lack isNormalized() copying? But I'd put this separately and still use _copy_metadata() as well

klemen1999 (commented):
For completness sake it might be good to also check other cases:
- FastSAM in [process_masks](https://github.com/luxonis/depthai-nodes/blob/53d61ade767f326a0efc0a39cff79e7d0833c370/depthai_nodes/node/parsers/utils/fastsam.py#L319)
- RF-DETR [here](https://github.com/luxonis/depthai-nodes/blob/53d61ade767f326a0efc0a39cff79e7d0833c370/depthai_nodes/node/parsers/utils/masks_utils.py#L83)
- Not sure if there are any others

depthai_nodes/message/utils/copy_message.py
klemen1999: Correction: Looks like dai.Point2f is not a dai.Buffer anyway so `_copy_metadata()` doesn't do anything to it anyway. And normalized attribute can only be set on the init so it has to be done the way you've done it right now

depthai_nodes/node/parsers/utils/masks_utils.py
dtronmans: Addressed in [edfefd6](https://github.com/luxonis/depthai-nodes/pull/329/commits/edfefd60160483959bfb5c76c33d2186d4bbb4d0)

depthai_nodes/node/parsers/utils/masks_utils.py
```diff
@@ -105,15 +122,11 @@ def process_single_mask_rfdetr(
...
-        mask,
+    _ = bbox
+    resized_mask_logits = cv2.resize(
+        mask_logits,
         (input_shape[1], input_shape[0]),
-        interpolation=cv2.INTER_NEAREST,
+        interpolation=cv2.INTER_LINEAR,
     )
+    logit_threshold = probability_to_logit_threshold(mask_conf)
+    return (resized_mask_logits > logit_threshold).astype(np.uint8)
```
klemen1999: Although the official RF-DETR repo doesn't perform cropping I would still have it in our implementation so it's the same between this and e.g. YOLO.

klemen1999: Note that crop has to happen after threshold for it to be correct

depthai_nodes/node/parsers/utils/masks_utils.py
```diff
@@ -46,14 +56,21 @@ def process_single_mask(
...
-    mask = crop_mask(mask, scaled_bbox)
-    return (mask > mask_conf).astype(np.uint8)
+    mask_logits = np.sum(protos * mask_coeff[..., np.newaxis, np.newaxis], axis=0)
+    mask_logits = crop_mask(mask_logits, scaled_bbox)
+    mask_logits = cv2.resize(
+        mask_logits,
+        (output_shape[1], output_shape[0]),
+        interpolation=cv2.INTER_LINEAR,
+    )
+    return (sigmoid(mask_logits) > mask_conf).astype(np.uint8)
```
klemen1999: We can probably use the same `probability_to_logit_threshold()` function instead of sigmoid here right? So:
`mask_logits > probability_to_logit_threshold(mask_conf)`

depthai_nodes/node/parsers/utils/masks_utils.py
```diff
@@ -46,14 +56,21 @@ def process_single_mask(
...
+    @return: Processed binary mask resized to `output_shape`.
     @rtype: np.ndarray
     """
     c, mh, mw = protos.shape  # CHW
     scaled_bbox = bbox * np.array([mw, mh, mw, mh])
-    mask = sigmoid(np.sum(protos * mask_coeff[..., np.newaxis, np.newaxis], axis=0))
-    mask = crop_mask(mask, scaled_bbox)
-    return (mask > mask_conf).astype(np.uint8)
+    mask_logits = np.sum(protos * mask_coeff[..., np.newaxis, np.newaxis], axis=0)
+    mask_logits = crop_mask(mask_logits, scaled_bbox)
```
klemen1999: Crop should happen on the binary mask as the last operation, after thresholding.
The way it's done now can lead to this mistake:
process_single_mask currently crops logits by setting out-of-bbox pixels to 0. But a logit of 0 means 50% confidence, not background. Therefore, with any allowed mask_conf < 0.5, those out-of-bbox pixels become foreground and can fill most of the mask.

dtronmans: Good catch, addressed in [0f8a7b4](https://github.com/luxonis/depthai-nodes/pull/329/commits/0f8a7b4ffa90a4b7cbe05d2c94c75c2eb29ec977)

depthai_nodes/node/parsers/utils/masks_utils.py
dtronmans: Addressed in [0f8a7b4](https://github.com/luxonis/depthai-nodes/pull/329/commits/0f8a7b4ffa90a4b7cbe05d2c94c75c2eb29ec977)

depthai_nodes/node/parsers/utils/masks_utils.py
dtronmans: Addressed in [0f8a7b4](https://github.com/luxonis/depthai-nodes/pull/329/commits/0f8a7b4ffa90a4b7cbe05d2c94c75c2eb29ec977)

klemen1999 (approved):

